PR #200, opened 2017-07-17
Add alerts to pages and global

hbeachey:
- Pages have alerts below their title
- Global alerts show up between logo and nav

### DIFF
--- a/src/components/Contentful/Alert/Global/index.js
+++ b/src/components/Contentful/Alert/Global/index.js
@@ -36,7 +36,7 @@ const mapStateToProps = (state) => {
   }
   return {
     allAlerts,
-    allAlertsStatus: state.allAlerts.status,
+    allAlertsStatus: state.allAlerts ? state.allAlerts.status : '',
   }
 }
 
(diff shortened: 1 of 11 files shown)


src/components/Contentful/Alert/Global/index.js
```diff
@@ -0,0 +1,73 @@
...
+import Presenter from '../presenter.js'
+import * as statuses from '../../../../constants/APIStatuses'
+import { flattenLocale } from '../../../../shared/ContentfulLibs'
+
+const mapStateToProps = (state) => {
+  let allAlerts = []
+  if (state.allAlerts && state.allAlerts.status === statuses.SUCCESS) {
+    let now = new Date()
+
+    allAlerts = state.allAlerts.json
```
jonhartzler: While this works it seems this is a case where it should be done with a selector.  
http://redux.js.org/docs/recipes/ComputingDerivedData.html

hbeachey: done

PR #204, opened 2017-07-19
Fix footer logo

danthewolfe:
* SVGs are handled differently by webpack than PNGs and were not being includedwhere expected. I converted the logo to PNG format so that it gets handled the same as other images.


hbeachey:
Is it better to convert the svg or to change where webpack puts the built file?

danthewolfe:
The svg handling was a default from CreateReactApp. I'm not entirely sure why lumping svg in with css and json was handled the way it was, but I figured since this is the only(?) image we have in svg, it's easier - at least for now - to convert the image than modify both the dev and production webpack configurations.

PR #208, opened 2017-07-20
Organize hours page

jonhartzler:
- Organized the hour's page off of a coded list.  
- Added a css field to allow for sub-service points to be listed as indented.


src/components/Hours/Page/presenter.js
```diff
@@ -8,6 +8,26 @@ import PageTitle from '../../PageTitle'
 import StaticSidebar from '../../Contentful/StaticContent/Sidebar'
 import StaticBody from '../../Contentful/StaticContent/Body'
 
+const hoursPageOrder = [
```
jgondron: Seems like you can just pass this in as part of the props from the higher component

PR #213, opened 2017-07-21
Alerts styling and layout fixes.

awetherill:


src/components/Courses/presenter.js
```diff
@@ -160,8 +160,9 @@ class Courses extends Component {
     return (
       <div className='container-fluid content-area'>
         <div key='courses' className='content'>
-          <Link to='/items-requests' className='button fright tab'>My Items</Link>
           <LogOut />
+          <Link to='/personal' className='button fright tab'>My Items</Link>
```
hbeachey: I think this link was changed for a reason @jonhartzler might know for sure

jonhartzler: Yes this should be /items-requests

PR #219, opened 2017-07-24
QA display fixes

awetherill:


danthewolfe:
And take your 30 pin connectors with you.

PR #227, opened 2017-07-27
Minor display updates.

awetherill:


src/components/LandingPages/Research/index.js
```diff
@@ -27,7 +27,7 @@ const Research = () => {
           </p>
           <p>
 
-            <Link to='https://directory.library.nd.edu/directory/subjects' noTarget={true}>Subject Librarians and Specialists</Link>
+            <Link to='https://directory.library.nd.edu/directory/subjects' noTarget={true}>Librarians by Subject &amp; Specialty</Link>
```
danthewolfe: I thought we recently made several changes for '&' to 'and'. Do we need to double check with @lajamie about the ampersand changes?

lajamie: RBSC's formal title has an & in it, so it needs it, and the & in Librarians by Subject/Specialty makes that line fit better in the space allotted. But we did indeed make a bunch of other changes from & to "and" on the subjects page to conform with university style guidelines and to make the page internally consistent.

PR #234, opened 2017-08-01
Bugfix timezone offsets

jgondron:
- The hours api now gives us the utcOffset for the times at each location. Changed the timeToday to use this information to convert all times to local times before doing comparisons.


jgondron:
Watched this overnight and seemed fine. Note: Depends on https://github.com/ndlib/monarch_libguides/pull/15 getting deployed first.

PR #258, opened 2017-08-14
Mh courses unit test

mhallwac:
Creates the test for the Courses/index.js file.  Uses the enzyme wrapper to test the Courses Container based on possible outcomes in the index.js file. Also adds export to the Courses Container class declaration because otherwise it cannot be tested.

Then, tests the mapStateToProps functionality  by using two props, "state" and "ownProp".  Tests with correct data and incorrect data.

### DIFF
--- a/src/tests/components/Courses/index.test.js
+++ b/src/tests/components/Courses/index.test.js
@@ -2,14 +2,14 @@ import * as statuses from '../../../constants/APIStatuses'
 import { mapStateToProps } from '../../../components/Courses'
 
 describe('components/Courses/index.js', () => {
-  describe('with data', () => {
+  describe('with correct data', () => {
     let state = {
       personal: {
         login: {
           state: statuses.SUCCESS,
         },
+        courses: ['courses']
       },
-      courses: ['courses']
     }
 
     let ownProp = {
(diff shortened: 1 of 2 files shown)


src/tests/components/Courses/index.test.js
```diff
@@ -0,0 +1,172 @@
...
+
+    it('does not render Courses Presenter', () => {
+      expect(enzymeWrapper
+        .containsMatchingElement(<CoursesPresenter {...props} />))
+        .toBe(false)
+    })
+
+    it('only renders a link', () => {
+      expect(enzymeWrapper
+        .containsMatchingElement(<Link to='/courses' className='button fright tab'>My Courses</Link>))
```
hbeachey: Will this test work without specifying the className? If so, prefer not to test css items (which is what className denotes in react)

src/tests/components/Courses/index.test.js
```diff
@@ -0,0 +1,172 @@
...
+          preview: 'true'
+        },
+      },
+    }
+
+    it('should identify if the user is logged in', () => {
+      expect(mapStateToProps(state, ownProp).loggedIn).toBe(true)
+    })
+
+    it('should be previewable', () => {
```
hbeachey: Generally in this project `preview` means "use unpublished content from contentful." So if `preview=false` it would only use published content (eg. production). Valid test, but would be more correct to be something like `it('should use preview content')`

PR #259, opened 2017-08-15
Add Search Scope to ND Catalog

danthewolfe:
* Added extra field to display on advanced search for ND Catalog
* Also added “ND Catalog” and “OneSearch” title to Advanced Search

### DIFF
--- a/src/components/SearchDrawer/AdvancedSearch/SearchScope/presenter.js
+++ b/src/components/SearchDrawer/AdvancedSearch/SearchScope/presenter.js
@@ -1,0 +1,31 @@
+import React from 'react'
+
+const SearchScope = (props) => {
+  if (props.searchType === 'NDCATALOG') {
+    return (
+      <div>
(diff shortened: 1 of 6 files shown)


src/components/SearchDrawer/AdvancedSearch/SearchScope/presenter.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react'
+
+const SearchScope = (props) => {
+  if (props.searchType === 'NDCATALOG') {
+    return (
+      <div>
+        <label htmlFor='exlidSearchIn'>Search Scope:</label>
+        <select id='scopesListAdvanced' name='scp.scps' onChange={props.onChange}>
```
hbeachey: This file feels like another thing that shouldn't be hard-coded. Don't know if that's really a priority though.

danthewolfe: There probably needs to be some refactoring done in advanced search code, but I think this is okay for now.

PR #265, opened 2017-08-17
Service now link

jgondron:
Created a component to handle service now links and moved the base url into the configuration so that we can more easily change this if the form url changes again.

### DIFF
--- a/src/components/ServiceNowLink/index.js
+++ b/src/components/ServiceNowLink/index.js
@@ -1,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Link from '../Link'
+import Config from '../../shared/Configuration'
+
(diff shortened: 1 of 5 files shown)


src/components/ServiceNowLink/index.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Link from '../Link'
+import Config from '../../shared/Configuration'
+
+const ServiceNowLink = ({ children, isWebContent }) => {
+  let url = `${Config.serviceNowBaseURL}&URL=${window.location}`
+  if (isWebContent) {
+    url += '&lib_list_problem=lib_list_web_content'
```
hbeachey: Should this also be in config?

PR #267, opened 2017-08-18
News on Homepage

hbeachey:
- show 3 most recent news entries on home page instead of timing out (sorted by starttime)
- fix node warnings
- fix css warnings

### DIFF
--- a/config/webpack.config.dev.js
+++ b/config/webpack.config.dev.js
@@ -170,7 +170,8 @@ module.exports = {
           'last 4 versions',
           'Firefox ESR',
           'not ie < 9' // React doesn't support IE8 anyway
-        ]
+        ],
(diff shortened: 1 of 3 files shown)


config/webpack.config.dev.js
```diff
@@ -170,7 +170,8 @@ module.exports = {
           'last 4 versions',
           'Firefox ESR',
           'not ie < 9' // React doesn't support IE8 anyway
-        ]
+        ],
+        grid: false,
```
danthewolfe: Does this fix the grid-row css warning in the dev console?

hbeachey: Yeah, I was afraid it might break IE, but this is just in dev so /shrug

PR #273, opened 2017-08-23
Renew single and Renew All

hbeachey:
- Add renew button to aleph items
- Add renew all button to top of checked out items
- Use custom action + reducer for reducers instead of the generic `personal` one

### DIFF
--- a/src/components/Hours/Current/expanded_presenter.js
+++ b/src/components/Hours/Current/expanded_presenter.js
@@ -26,11 +26,11 @@ const Presenter = (hoursEntry, isOpen, collapseHandler, children) => {
         </h4>
       </a>
       <div className='row hours-listing' id={hoursEntry.servicePoint.slug}>
-        <div className='col-md-4'>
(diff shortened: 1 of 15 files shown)


src/components/Hours/Current/expanded_presenter.js
```diff
@@ -26,11 +26,11 @@ const Presenter = (hoursEntry, isOpen, collapseHandler, children) => {
         </h4>
       </a>
       <div className='row hours-listing' id={hoursEntry.servicePoint.slug}>
-        <div className='col-md-4'>
```
jgondron: I have a similar change Andy and I worked through here https://github.com/ndlib/usurper/pull/274/files#diff-fe06ef3670400d83e4d88e867692457fL29. Just double check the values to see which one is correct. You'll prob get a conflict.

hbeachey: Andy made this change, so /shrug

PR #280, opened 2017-08-30
Aleph Retry

hbeachey:
- Retry queries that fail when getting information from aleph and illiad.
- Split up borrowed/pending queries into aleph/illiad + borrowed/pending. 
  - This means the `/items-requests` page now does 4 queries to get items + 1 for user information (5 total)
  - Requires [this gatekeeper pr](https://github.com/ndlib/gatekeeper/pull/22) to be merged
- Attempt to make page more friendly by displaying loaded items along with loading bar when aleph takes long time


src/actions/personal/loanResources.js
```diff
@@ -36,42 +12,81 @@ export const handleResources = (dispatch, data) => {
...
+
+const doQuery = (dispatch, service, type, func, token, stateKey, retry = 0) => {
+  dispatch(states.requestPersonal(stateKey))
+  states.startRequest(
+    Config.resourcesAPI + '/' + service + '?type=' + type,
+    dispatch,
+    func,
+    token,
+    (e) => {
+      console.log(e)
```
danthewolfe: Are these console.logs useful to keep?

hbeachey: Yeah, these are the error handlers, printing the exceptions. The retry message I think is also good to keep, it's kinda debuggy but for users savvy enough to look at the console it would still be helpful.

PR #283, opened 2017-09-07
Microdata

hbeachey:
Starting to add microdata to news/events/contact info.

News/events require more work and more fields to be added to contentful, but until we decide what's happening with them I'm going to hold off.
- Events need location.
- News needs: Author, datepublished, and publisher

Changes tested with https://search.google.com/structured-data/testing-tool

### DIFF
--- a/src/components/LandingPages/Events/presenter.js
+++ b/src/components/LandingPages/Events/presenter.js
@@ -5,24 +5,35 @@ import PageTitle from '../../PageTitle'
 import SearchProgramaticSet from '../../SearchProgramaticSet'
 import './style.css'
 
-const makeEntry = (entry) => {
+export const makeEventEntry = (entry, index) => {
   return (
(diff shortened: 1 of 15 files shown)


src/components/LandingPages/Events/presenter.js
```diff
@@ -5,24 +5,35 @@ import PageTitle from '../../PageTitle'
...
+      itemScope
+      itemType='http://schema.org/Event'
+      key={'events_' + index}
+    >
+      <Link
+        ariaLabel={entry.title + ' on ' + entry.displayWeekday + ', ' + entry.displayDay + ' ' + entry.displayMonth + ' ' + entry.displayYear}
+        to={'/event/' + entry.slug}
+        itemProp='mainEntityOfPage'
+      >
+        <time dateTime={entry.startDate} className='date-as-calendar inline-flex'>
```
danthewolfe: Is this affected by user timezone?

hbeachey: Likely yes, the startdate in contentful has timezone though so I think it should be correct either way.

danthewolfe: :+1:

PR #284, opened 2017-09-11
LW-91 Update hours api

jonhartzler:
Updated to use the new monarch libguides api endpoint. 

1.  moved the collapsing of similar hours from the api to this item 
2.  changed the effective date calculation to say starting at
3.  changed the format of the data to match the new api end point.
4.  changed the way that the calculation of what is open works it uses the date with the timezone to make the calculation.


config/configurationGen.js
```diff
@@ -7,7 +7,7 @@ const genConfig = () => {
     coursesAPI: 'https://unauxwjiml.execute-api.us-east-1.amazonaws.com/dev',
     resourcesAPI: 'https://p5gotxkf09.execute-api.us-east-1.amazonaws.com/dev',
     illiadBaseURL: 'https://nd.illiad.oclc.org/illiad/IND/testweb/illiad.dll?Action=10&Form=<<form>>&Value=<<value>>',
-    hoursAPIURL: 'https://3nmbw29969.execute-api.us-east-1.amazonaws.com/dev',
```
hbeachey: Undo!

config/configurationGen.js
```diff
@@ -7,7 +7,7 @@ const genConfig = () => {
...
+    hoursAPIURL: 'https://qqmk3lr1a9.execute-api.us-east-1.amazonaws.com/dev',
```
hbeachey: I think this should still be `3nm...` the `qqm..` prefix is still `devJon`

src/components/Hours/HomePage/index.js
```diff
@@ -17,7 +17,7 @@ const makeMapStateToProps = () => {
     let props = {
       servicePoint: {
         fields: {
-          hoursCode: 'hesburghlibrary',
+          hoursCode: '426',
```
hbeachey: Should probably either make this a const define, or at least add a comment.

config/configurationGen.js
```diff
@@ -7,7 +7,7 @@ const genConfig = () => {
     coursesAPI: 'https://unauxwjiml.execute-api.us-east-1.amazonaws.com/dev',
     resourcesAPI: 'https://p5gotxkf09.execute-api.us-east-1.amazonaws.com/dev',
     illiadBaseURL: 'https://nd.illiad.oclc.org/illiad/IND/testweb/illiad.dll?Action=10&Form=<<form>>&Value=<<value>>',
-    hoursAPIURL: 'https://3nmbw29969.execute-api.us-east-1.amazonaws.com/dev',
+    hoursAPIURL: 'https://qqmk3lr1a9.execute-api.us-east-1.amazonaws.com/devJon',
```
hbeachey: .

src/components/Hours/Current/collapsed_presenter.js
```diff
@@ -2,14 +2,21 @@ import React from 'react'
...
 import WeeklyHours from '../WeeklyHours'
+const days = ['su', 'mo', 'tu', 'we', 'th', 'fr', 'sa']
 
 const Presenter = (hoursEntry, isOpen, expandHandler, children) => {
   const title = 'Hours for ' + hoursEntry.name
   const servicePointClassName = 'service-point ' + (isOpen ? 'open' : 'closed')
-  const todayLabel = 'Today: ' + hoursEntry.today.display
+  const todayLabel = 'Today: ' + hoursEntry.today.rendered
 
+  const schemaOpeningHours = days[new Date().getDay()] + ' ' + hoursEntry.today.rendered
```
hbeachey: I'd take a look at how I did this in events `.../Home/Events/index.js` the date class has a `toLocaleString` that can remove the `days` and `months` arrays throughout this PR.

jonhartzler: The main issue I have is that toLocaleString does not support the 2 letter days format schema.org needs.   

The representation of the weekday. Possible values are "narrow", "short", "long".

narrow is a single letter and short is the 3 letter.  
d.toLocaleString('en-US', {weekday: 'narrow'})
"W"

What I did do was move the calculation to the selector so there is less duplication.

PR #285, opened 2017-09-11
LW-72 Design remediation for items page

danthewolfe:
* Title column  wider
* Date column narrower
* Placed permanent down/up arrows over each column
* Changed “More Actions” button to tricolor instead of hamburger and
added aria-label and title attribute.
* Moved “Renew All” button  to make “More Actions” column narrower.


src/components/LoanResources/ResourceList/presenter.js
```diff
@@ -8,6 +8,7 @@ const ResourceList = (props) => {
     <div>
       <div className='filter'>
         Filter items: <input type='text' value={props.filterValue} onChange={props.filterChange} />
+        { props.borrowed && <button className='renew' onClick={props.renewAll} aria-label='Renew All'>Renew All</button> }
```
jonhartzler: Just a couple thoughts on accessibility here.  

1.  aria-label is not needed when the labels match.   
2.  That said, I think the aria-label should be more descriptive because it could be read out of context.  Maybe something like "Renew all overdue items" . or Renew all Renewable items!!   I am not sure maybe you can talk to aaron.

PR #286, opened 2017-09-11
Database search

hbeachey:
Add database search box to databases a-z page. 
- Query for all databases on load of any a-z page
- Search all databases by title
- Search results limited to 50

Also moved some logic out of presenter and into the container/mapping functions.

### DIFF
--- a/src/components/DatabaseList/presenter.js
+++ b/src/components/DatabaseList/presenter.js
@@ -5,15 +5,17 @@ import '../../static/css/global.css'
 import PageTitle from '../PageTitle'
 import SearchProgramaticSet from '../SearchProgramaticSet'
 import Link from '../Link'
-import PageNotFound from '../Messages/NotFound'
 import ErrorLoading from '../Messages/Error'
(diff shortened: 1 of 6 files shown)


src/components/DatabaseList/presenter.js
```diff
@@ -5,15 +5,17 @@ import '../../static/css/global.css'
...
 import * as statuses from '../../constants/APIStatuses'
 
-const Content = (letter, data) => {
+const Content = (letter, data, filterValue, onFilterChange) => {
   return (
     <div className='container-fluid content-area'>
       <PageTitle title={'Databases: ' + letter.toUpperCase()} />
       <SearchProgramaticSet open={false} />
+      <div className='filter'>
+        <strong>Search all Databases by Title: </strong><input type='text' value={filterValue} onChange={onFilterChange} />
```
jonhartzler: Please add a label tag for accessibility.  

We will continue to have the problem that searching changes the page and we are not labeling that change correctly.  Please ticket that.

PR #300, opened 2017-10-10
Markdown tables

hbeachey:
style tables to be more grid-like - much nicer!

### DIFF
--- a/src/components/LibMarkdown/index.js
+++ b/src/components/LibMarkdown/index.js
@@ -52,10 +52,33 @@ class LibMarkdown extends Component {
       return null
     }
 
+    const toDiv = (className) => { return (props) => (<div className={className} {...props.props}>{props.children}</div>) }
+
     let overrides = {
       overrides: {
         a: {
           component: LibLink,
+        },
+        tr: {
+          component: toDiv('markdown-tr'),
+        },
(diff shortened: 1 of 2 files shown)


src/components/LibMarkdown/index.js
```diff
@@ -52,11 +52,34 @@ class LibMarkdown extends Component {
       return null
     }
 
+    const toDiv = (className) => { return (props) => (<div className={className} {...props.props}>{props.children}</div>) }
```
danthewolfe: Why do we need to transform the html here? Why not just redefine the appearance in css?

If users are trying to use tables as a layout tool I'm fine with these changes. But if we just want to alter the appearance of actual tabular data I'd prefer maintaining the table related tags for their semantic value and just restyling to get away from the "table" appearance.

hbeachey: Fair point, I'm not really sure what the best option is - or if I went the right direction at all. 

I was trying to make this fit the same styling as the rest of the site - using the grid layout. Granted, genericizing a grid layout to look good for an arbitrary number of columns is not trivial, so perhaps it would be best to just mess with the table css. I can go back and try to do something similar, keeping the table tags.

PR #301, opened 2017-10-12
Services Landing Page.

danthewolfe:


hbeachey:
I left a note on the ticket. Seems rather cramped. The services page is good, but the pages that have  more than 1 column are pretty cramped.

awetherill:
I think the intention was to have all of these landing pages be one column with a navigational box on the right side to jump to each section (shown in the static mockup). I'll have to go in and adjust for mobile view as well.

PR #302, opened 2017-10-18
Update nav for accessability

jonhartzler:
Simple updates

1.  Added some additional aria to follow standards.
2.  Updated the html to make use of lists
3.  Added esc as a key to close the menu 
4.  Removed the idea that clicking on the link would go to a sub page.


src/components/Navigation/index.js
```diff
@@ -73,9 +68,8 @@ const mergeProps = (state, dispatchProps, ownProps) => {
 
   // Treat menu dropdowns as links to landing pages for accessibiilty
   const keyDown = (e) => {
-    if (e.keyCode === 13) {
-      const id = e.target.firstChild.id
-      ownProps.history.push(`/${id}`)
+    if (e.keyCode === 27) {
```
hbeachey: Add a comment of what key this is for posterity

src/components/Navigation/index.js
```diff
@@ -73,9 +68,8 @@ const mergeProps = (state, dispatchProps, ownProps) => {
 
   // Treat menu dropdowns as links to landing pages for accessibiilty
```
danthewolfe: remove comment since it's doing something else now.

PR #308, opened 2017-10-25
INC0149167

danthewolfe:
Only include dates if they are set by the user. This is a partial fix
to dates incorrectly being added to the second and third free text
fields.

### DIFF
--- a/src/components/SearchDrawer/searchQueryBuilder.js
+++ b/src/components/SearchDrawer/searchQueryBuilder.js
@@ -38,8 +38,8 @@ const searchQuery = (searchStore, advancedSearch, history) => {
     const drStartDay = advancedSearch['drStartDay'] || '01'
     const drStartMonth = advancedSearch['drStartMonth'] || '01'
     let drStartYear = advancedSearch['drStartYear5']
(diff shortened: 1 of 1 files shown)


src/components/SearchDrawer/searchQueryBuilder.js
```diff
@@ -35,12 +35,12 @@ const searchQuery = (searchStore, advancedSearch, history) => {
     const bool1 = advancedSearch['bool_1'] || 'AND'
     const materialType = advancedSearch['materialType'] || 'all_items'
     const language = advancedSearch['language'] || 'all_items'
-    const drStartDay = advancedSearch['drStartDay'] || '00'
-    const drStartMonth = advancedSearch['drStartMonth'] || '00'
-    const drStartYear = advancedSearch['drStartYear5'] || '1900'
+    const drStartDay = advancedSearch['drStartDay'] || '01'
```
hbeachey: Does it make sense for the start day to default to `01` and the end day to default to `00`? (same for month).

danthewolfe: Adjusted end date to December 31.

PR #323, opened 2017-10-31
all landing pages are 1 column now

hbeachey:


hbeachey:
We can do that @rfox2 It was implemented slightly confusingly, as it would only be 1 column if there were more than 2 columns in contentful. We can change it to be 1 if there is only 1 in contentful if that's what we'd like instead.

PR #326, opened 2017-11-02
LW-161 Replace CSS "fixed" position with React-based positioning

danthewolfe:
CSS alone can’t get this to behave the way we want.

### DIFF
--- a/src/components/Contentful/Page/sideNavPresenter.js
+++ b/src/components/Contentful/Page/sideNavPresenter.js
@@ -1,24 +1,75 @@
-import React from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+const SCROLLED = true
+const NOTSCROLLED = false
+const MAGICNUMBER = 245
+class SideNavPresenter extends Component {
+  constructor (props) {
+    super(props)
+    this.scrolledStyle = this.scrolledStyle.bind(this)
+    this.listenScrollEvent = this.listenScrollEvent.bind(this)
+    this.state = {
+      positionType: NOTSCROLLED,
+    }
+  }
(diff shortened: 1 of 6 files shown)


src/components/Contentful/Page/sideNavPresenter.js
```diff
@@ -1,24 +1,75 @@
-import React from 'react'
+import React, { Component } from 'react'
 import PropTypes from 'prop-types'
+const SCROLLED = true
+const NOTSCROLLED = false
+const MAGICNUMBER = 245
```
hbeachey: Solid. If you call it a magic number is it a magic number? 🤔

src/components/Contentful/Page/sideNavPresenter.js
```diff
@@ -1,24 +1,75 @@
...
+    if (window.pageYOffset <= MAGICNUMBER && this.state.positionType === SCROLLED) {
+      this.setState({ positionType: NOTSCROLLED })
+    }
+  }
+
+  scrolledStyle () {
+    if (this.state.positionType === SCROLLED) {
+      return {
+        position: 'fixed',
+        top: '10px',
```
hbeachey: If the search bar is down and advanced search open this overlaps quite a bit.

src/components/Contentful/Page/sideNavPresenter.js
danthewolfe: :trollface:

src/components/Contentful/Page/sideNavPresenter.js
danthewolfe: This should be fixed now.

PR #328, opened 2017-11-06
Accesability changes

jonhartzler:
A number of accessibility changes.

1.  Changed the way that page sections are done to be more html5 compliant
2.  Changed the way that hours and the chat tab announce to screen readers.
3.  Changed the page title to an h1.
4.  Changed some of the h2 - h6
5.  Changed the Database search page so that the next item after the search in the keyboard focus is the search results. 
6.  changed the way that the search dropdowns role.


src/static/css/global.css
```diff
@@ -603,7 +629,7 @@ body.gold-blue #header h4 a {
   padding:9px 11px !important;
 }
 
-#banner .logout {
+/*#banner .logout {
```
hbeachey: Maybe just delete this?

PR #332, opened 2017-11-08
Event landing

hbeachey:
Clean up the events landing page a bit
- Clean up styling of event list (affects home page as well)
- Landing now only shows `Current Events`, with a toggle button between that and `Past Events`
- Show image next to events
  - Required altering the image component to grab images if they weren't already in the CF data
  - Image change also affected (slightly) the database list
  - Event list + Database List will now only load 1 time instead of every time the component mounts


hbeachey:
Closing until Tara's team signs off

PR #338, opened 2017-11-30
Color Changed to Pass Contrast Test

Nathan-Ladwig:
Changed the colors on the hours and Spanish page to pass the color contrast test.

### DIFF
--- a/src/components/Hours/Current/collapsed_presenter.js
+++ b/src/components/Hours/Current/collapsed_presenter.js
@@ -33,7 +33,7 @@ const Presenter = (hoursEntry, isOpen, expandHandler, children) => {
           </div>
         </h4>
       </a>
-      <div id={hoursEntry.servicePoint.slug} role='tabpanel' aria-hidden={true} />
+      <div className={hoursEntry.servicePoint.slug} role='tabpanel' aria-hidden={true} />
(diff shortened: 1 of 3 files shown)


src/components/Hours/Current/collapsed_presenter.js
```diff
@@ -33,7 +33,7 @@ const Presenter = (hoursEntry, isOpen, expandHandler, children) => {
           </div>
         </h4>
       </a>
-      <div id={hoursEntry.servicePoint.slug} role='tabpanel' aria-hidden={true} />
+      <div className={hoursEntry.servicePoint.slug} role='tabpanel' aria-hidden={true} />
```
hbeachey: This needs to be `id` so that the `aria-controls` in the div above will be able to find it correctly. Unless I'm mistaken, will this still work for screen readers?

Nathan-Ladwig: We tested voice controls using both id and classname, and it appears to work exactly the same expanded and collapsed.

hbeachey: @jonhartzler ^ to pass along knowledge. I didn't realize `className` would work for `aria-controls`

PR #340, opened 2017-12-06
Changed ARIA role to ID

Nathan-Ladwig:
Stops validation error for ARIA roles if we change 'role' in front of copyright to 'id'.

@jonhartzler

### DIFF
--- a/src/components/Footer/FooterInfo/index.js
+++ b/src/components/Footer/FooterInfo/index.js
@@ -11,7 +11,11 @@ const FooterInfo = () => {
         <div className='row'>
           <div className='col-xs-12 col-md-4'>
             <div className='box'>
-              <p role='copyright'><Link to='https://www.nd.edu/copyright/'>Copyright</Link> © 2017 University of Notre Dame</p>
+              <p id='copyright' property='dc:rights'>
+                <Link to='https://www.nd.edu/copyright/'>Copyright</Link>&nbsp;
+                <span property='dc:dateCopyrighted'>© 2017</span>&nbsp;
+                <span property='dc:publisher'>University of Notre Dame</span>
+              </p>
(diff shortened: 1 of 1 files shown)


src/components/Footer/FooterInfo/index.js
```diff
@@ -11,7 +11,11 @@ const FooterInfo = () => {
         <div className='row'>
           <div className='col-xs-12 col-md-4'>
             <div className='box'>
-              <p role='copyright'><Link to='https://www.nd.edu/copyright/'>Copyright</Link> © 2017 University of Notre Dame</p>
+              <p id='copyright' property='dc:rights'>
```
danthewolfe: @jonhartzler I'm not sure we want to go Dublin Core here, but since there is no aria role for copyright, this seemed like a reasonable alternative. I referred @Nathan-Ladwig to this post: https://stackoverflow.com/questions/254613/semantic-html-markup-for-a-copyright-notice

I'm not sure if DC needs `xmlns:dc="http://purl.org/dc/elements/1.1/"` in a parent element, but the namespaced property `xmlns:dc` does not play well with React (so that might be a reason not to go this route).

jonhartzler: I think the solution is that role=copyright should just be removed and not replaced with metadata.  role=copyright is a proposed standard.  I imagine I used it by referencing an incorrect list of possible roles.   The inclusion of the metadata schema will not fix accessibility as those browsers don't tend to read it also it is not data we care to actually push.  Including this would only be for the search engine itself and I also am not certain that this is the metadata we want to send to the search engine.  Since it will prioritize it and we don't have metadata about large sections of our actual content.

danthewolfe: @jonhartzler @Nathan-Ladwig 
So revert changes except: 

`role='copyright'` -> `id='copyright'`

jonhartzler: yeah I think so

Nathan-Ladwig: Reapplied another pull request with just the change from role to id.

PR #343, opened 2017-12-12
Event landing

awetherill:


src/components/Contentful/Database/index.js
```diff
@@ -6,20 +6,27 @@ import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import PresenterFactory from '../../APIPresenterFactory'
 import ContentfulDatabasePresenter from './presenter.js'
+import * as statuses from '../../../constants/APIStatuses'
 
-const mapStateToProps = (state) => {
-  return { cfDatabaseEntry: state.cfEntry }
+const mapStateToProps = (state, thisProps) => {
+  const databaseId = thisProps.match.params.id
```
jonhartzler: This looks duplicated in this location and in the container below.  Should that be?

src/components/Contentful/Event/presenter.js
```diff
@@ -22,12 +24,16 @@ const PagePresenter = ({ entry }) => (
...
+        <h3>Date &amp; Time</h3>
+        <p>{entry.fields.hoursDisplay}</p>
         <LibMarkdown itemProp='description'>{ entry.fields.content }</LibMarkdown>
         <Related className='p-resources' title='Resources' showImages={false}>{ entry.fields.relatedResources }</Related>
       </main>
       <asside className='col-md-4 right'>
         <Image cfImage={entry.fields.representationalImage} className='cover' />
+        <Link to={entry.fields.registrationUrl} className='button callout' hideIfNull>Register Here</Link>
         <Librarians netids={entry.fields.contactPeople} />
+        <ServicePoint cfServicePoint={entry.fields.location} showHours={false} />
```
jonhartzler: We choose not to show the hours on the Event Page?  It seems like we want consistency?

hbeachey: If the event has something like "Hours: 8AM-11PM" but if the location hours are "Opens at Noon" then that's confusing, is the event happening or not? So we decided to only show the event hours on those pages.

src/components/Contentful/Database/index.js
hbeachey: Fixed!

PR #352, opened 2017-12-19
LW-235 BUG - Hours wrapping issue

danthewolfe:
See SN ticket:
https://nd.service-now.com/nav_to.do?uri=incident.do%3Fsys_id=5cb3a6d5db
0f03007314fce9af961948%26sysparm_stack=incident_list.do%3Fsysparm_query=
active=true

* Fix positioning of long hours
* Fix padding issue on individual pages, but leave in place for hours
page.

### DIFF
--- a/src/components/Hours/WeeklyHours/presenter.js
+++ b/src/components/Hours/WeeklyHours/presenter.js
@@ -17,7 +17,7 @@ const Presenter = ({ hours, title, effectiveDate, showEffectiveDates }) => {
         {
           hours.map((row) => {
             return (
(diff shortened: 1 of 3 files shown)


src/components/Hours/WeeklyHours/presenter.js
```diff
@@ -17,9 +17,9 @@ const Presenter = ({ hours, title, effectiveDate, showEffectiveDates }) => {
         {
           hours.map((row) => {
             return (
-              <span key={row.title}>
+              <span key={row.title} style={{ float: 'left', clear: 'left', width: '100%' }}>
```
jonhartzler: Is there a reason that we want these styles to run inline?

PR #362, opened 2018-01-08
Update react to 16.2.0

danthewolfe:


danthewolfe:
See changes to enzyme 3: https://github.com/ndlib/usurper/pull/362

hbeachey:
Still approved.

There's no way to only have to `configure(...)` one time instead of in every test file?

PR #364, opened 2018-01-10
display multiple entry points on db list page

rdoughty:


src/components/DatabaseList/presenter.js
```diff
@@ -66,17 +67,39 @@ const Loaded = (props) => {
...
     return (
-      <div key={item.fields.alephSystemNumber + item.fields.title} aria-label={item.fields.title} className='dbSection'>
-        <Link to={item.fields.purl} title={'Go to ' + item.fields.title}>
-          <h2 className='dbItem'>{item.fields.title}</h2>
-        </Link>
+      <section key={item.fields.alephSystemNumber + item.fields.title}
+        aria-label={item.fields.title} className='dbSection'>
+        {
+          exactlyOneUrl ? (
+            <Link to={item.fields.purl} title={'Go to ' + item.fields.title}>
```
hbeachey: Could you make this use the first url in the list instead of `fields.purl` I think we'd like to get rid of the `purl` field in the future.

If the base more info page is also using `purl` please change that as well

rdoughty:
These changes have been made and are ready for re-review :)

src/components/DatabaseList/presenter.js
```diff
@@ -66,17 +67,40 @@ const Loaded = (props) => {
...
+        {
+          exactlyOneUrl ? (
+            <Link to={item.fields.urls[0].url} title={'Go to ' + item.fields.title}>
+              <h2 className='dbItem'>{item.fields.title}</h2>
+            </Link>
+          ) : (
+            <h2 className='dbItem'>{item.fields.title}</h2>
+          )
+        }
+        <ul className='clamp databaseLink'>
```
hbeachey: @jonhartzler does an empty list mess with accessibility at all?

jonhartzler: no it should be fine.  screen readers will skip empty tags just like the browser.

PR #365, opened 2018-01-12
LW246 - contentful person and event migration

rdoughty:


deploy/contentfulMigrations/2018.01.11.addPerson.js
```diff
@@ -0,0 +1,62 @@
...
+  person.createField('phone')
+    .type('Symbol')
+    .name('Phone')
+
+  person.createField('picture')
+    .type('Link')
+    .name('Picture')
+    .linkType('Asset')
+
+  person.createField('type')
```
hbeachey: We might want to call this "Presenter Type" to make it more clear, though I suppose we may add non-presenter types to this in the future. IDK

rdoughty:
reworking event person mapping in contentful.  will have event --> presenter --> person

PR #366, opened 2018-01-15
lw-255 - date, time, and location overrides for contentful events

rdoughty:
https://jira.library.nd.edu/browse/LW-255


deploy/contentfulMigrations/2018.01.15.addLibCalEventTimeLocFields.js
```diff
@@ -0,0 +1,23 @@
+const forward = (migration) => {
+  const event = migration.editContentType('event')
+  event.createField('locationOverride')
+    .type('Symbol')
+    .name('Location Override')
+
+    event.createField('dateOverride')
+    .type('Symbol')
```
hbeachey: Can we make and `timeOverride` `text` instead of `symbol` so they have more space to override. Also, as discussed I don't think we need the `dateOverride` field.

PR #368, opened 2018-01-16
LW-259 - additional padding after last event

rdoughty:
https://jira.library.nd.edu/browse/LW-259

lw259_final1


src/components/Home/Events/presenter.js
```diff
@@ -5,12 +5,14 @@ import LibMarkdown from '../../LibMarkdown'
 import Image from '../../Image'
 
 export const makeEventEntry = (entry, index, isLast = false, showDescription = true, showImage = true) => {
+  const lastEventPadding = isLast === true ? { marginBottom: '36px' } : {}
```
danthewolfe: Do we want to do this in React or stick with a pseudo element in css?

`.event-card:last-of-type { marginBottom: 36px;}`

rdoughty:
modified to use last-of-type

PR #386, opened 2018-01-31
Lw261 - event locations and time changes

rdoughty:


src/components/Contentful/Event/presenter.js
```diff
@@ -12,49 +12,54 @@ import SearchProgramaticSet from '../../SearchProgramaticSet'
...
-      <aside className='col-md-4 right'>
-        <Image cfImage={entry.fields.representationalImage} className='cover' />
-        <Link to={entry.fields.registrationUrl} className='button callout' hideIfNull>Register Here</Link>
-        <Librarians netids={entry.fields.contactPeople} />
-        <ServicePoint cfServicePoint={entry.fields.location} showHours={false} />
-        <Related className='p-pages' title='Related Pages' showImages={false}>{ entry.fields.relatedPages }</Related>
-      </aside>
-    </div>
-  </article>
+  const defaultTime = `${hour12(new Date(entry.fields.startDate))} - ${hour12(new Date(entry.fields.endDate))}`
```
hbeachey: Are the `startDate` and `endDate` not the `Date` type already?

src/components/Contentful/Event/presenter.js
```diff
@@ -12,49 +12,54 @@ import SearchProgramaticSet from '../../SearchProgramaticSet'
...
+  const eventTime = entry.fields.timeOverride !== undefined ? entry.fields.timeOverride : defaultTime
```
hbeachey: I would recommend just doing `entry.fields.timeOverride ? .. : .. ` so that empty strings and null will be caught as well.

PR #390, opened 2018-01-31
LW-265: As a user of the libraries' website, I want to have images an…

danthewolfe:
…d more details shared with my post on Facebook

Create an OpenGraph component that adds meta tags to pages and include on Contentful pages, news, and events.
Based on info from : https://developers.facebook.com/docs/sharing/webmasters#markup

Note: Adds react-helmet module. Please run `yarn install` to add.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
     "react-ga": "^2.2.0",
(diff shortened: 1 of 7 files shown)


danthewolfe:
@jonhartzler By default, I don't think so. There is an additional metatag that can be added `<meta property='fb:app_id' content='[FACEBOOK_APP_ID]' />`.

PR #421, opened 2018-02-21
Display media links on events

hbeachey:


hbeachey:
@awetherill I copied the link sizing/padding from the map/contact info stuff. Is that what we want or should these links be smaller like the related pages?
screen shot 2018-02-21 at 10 58 16 am

PR #431, opened 2018-03-06
Fix a couple of console warnings

danthewolfe:
Fixes:

 * Warning: Unknown props `location`, `computedMatch` on \<div\> tag. Remove these props from the element.

 * Warning: Failed prop type: The prop `loggedIn` is marked as required in `Navigation`, but its value is `undefined`.


src/components/Navigation/presenter.js
```diff
@@ -107,7 +107,7 @@ Navigation.propTypes = {
   toggleClass: PropTypes.string.isRequired,
   handleMobileMenuKeyDown: PropTypes.func.isRequired,
   handleMobileClick: PropTypes.func.isRequired,
-  loggedIn: PropTypes.bool.isRequired,
+  loggedIn: PropTypes.bool,
```
danthewolfe: `undefined` is a valid and checked for.

PR #432, opened 2018-03-06
Add search console links.

jonhartzler:


public/index.html
```diff
@@ -10,6 +10,18 @@
     })(window,document,'script','dataLayer','GTM-TWL3G7T');</script>
     <!-- End Google Tag Manager -->
 
+    <!-- GA -->
+    <script>
```
danthewolfe: Can we do this with ReactGA instead? We already have the module loaded to do tracking on the search box.

See: https://github.com/ndlib/zucchetto/blob/d8a08d42baa41a8074016f21590e846c078cdd49/src/components/Analytics.jsx

and 

https://github.com/ndlib/zucchetto/blob/d8a08d42baa41a8074016f21590e846c078cdd49/src/routes/routes.js

for how I added it on Convocate.

PR #436, opened 2018-03-12
LW-349: Users need a calendar picker for Events

danthewolfe:
As described in LW-293:

* Adds calendar widget to current and past events landing pages.
* Create new landing page and add route /events/:YYYYMMDD and to show only events occuring on that date (includes multiday events with start dates before and end dates after selected date)
* Calendar widget "highlights" days that have an event on them, with the exception of events lasting more than 28 days - Highlighting every day of a month-long or longer events like Special exhibits makes finding other events occuring in that time period very difficult. Events that last longer than 28 days have their start and end date highlighted. Even though they are not highlighted on the calendar widget, they still appear in the list of events for a particular date.


src/components/LandingPages/Events/Calendar/index.js
```diff
@@ -0,0 +1,74 @@
...
+      const end = moment(event.endDate).format('YYYYMMDD')
+      let current = start
+      while (current <= end) {
+        specialDays.push({ date: moment(current, 'YYYYMMDD') })
+        current = moment(current, 'YYYYMMDD').add(1, 'day').format('YYYYMMDD')
+      }
+    }
+  })
+
+  const theme = {
```
hbeachey: Is overriding these in css not possible?

src/components/LandingPages/Events/Date/presenter.js
```diff
@@ -0,0 +1,55 @@
...
+import '../style.css'
+import { makeEventEntry } from '../../../Home/Events/presenter'
+import Link from '../../../Link'
+import Calendar from '../Calendar'
+import { withRouter } from 'react-router'
+import moment from 'moment'
+
+const EventDate = (props) => {
+  // For PageTitle
+  const displayDate = moment(props.match.params.date).format('MMM Do YYYY')
```
hbeachey: Check to see if date is an actual date, putting something else there (like `test`) causes an infinite loading bar instead of an error page.

danthewolfe: I've added a validation check on `props.match.params.date`. Any invalid date will redirect back to /events

src/components/LandingPages/Events/Calendar/index.js
danthewolfe: Yes, it is possible to change colors in css. I'd prefer if the styling were close to the component and not in the global stylesheet. 

Also, since the module author exposed styling in this way I'm inclined to go with it to hopefully prevent having to restyle if there are internal changes to the module.

hbeachey: 👍

danthewolfe:
Fixed the selected date not highlighting issue: https://github.com/ndlib/react-date-range/pull/1/files

danthewolfe:
Passed UA.

PR #443, opened 2018-03-22
LW 380 & LW366

danthewolfe:
**LW380:** Users want to see a headline on an individual news story that extends to the full width of the column.

- Add css wrapping class for news. Add css style to set wrapped h1 to 100% width.

**LW366:** Users need to see an en dash between between start and end times on individual events.

- Change regular hyphen/dash to `&ndash;`

### DIFF
--- a/src/components/Contentful/News/style.css
+++ b/src/components/Contentful/News/style.css
@@ -1,3 +1,6 @@
 .headline {
 	max-width: 30ch;
 }
+.news-article h1 {
+  max-width: 100%;
+}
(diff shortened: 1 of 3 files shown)


src/components/Contentful/News/style.css
```diff
@@ -1,3 +1,3 @@
-.headline {
-	max-width: 30ch;
-}
\ No newline at end of file
+.news-article h1 {
+  max-width: 100%;
```
hbeachey: We may want to keep this at the same max char width as the `p` tags to keep with accessibility. I'm not sure if that applies to `h` tags though.

PR #459, opened 2018-04-24
LW-406 Address page load speeds of Database A-Z pages on on non-webkit browsers

danthewolfe:
* Added `shouldComponentUpdate` method to the DatabaseList component to reduce number of redundant `render` calls.
* Removed `react-text-ellipsis` component.
* Added `-webkit` CSS to handle ellipsis.
* Added 'Loading' component during FETCHING cycle as UI enhancement.
* Fixed tests to reflect changes in adding 'Loading' component.

**REGRESSION:**
* Ellipsis no longer appear on Microsoft browsers or Firefix (however page load speed is greatly increased)

**NOTE:**
* `react-text-ellipsis` does browser detection on each usage. If webkit was detected, ellipsis was handled with CSS. If a non-webkit browser is detected an event listener is added on each occurrence and the text that is passed as a child is manipulated through javascript.

### DIFF
--- a/src/components/DatabaseList/presenter.js
+++ b/src/components/DatabaseList/presenter.js
@@ -11,7 +11,6 @@ import FilterBox from '../FilterBox'
 import OpenGraph from '../OpenGraph'
 import SideNav from '../SideNav'
 import LibMarkdown from '../LibMarkdown'
-import TextEllipsis from 'react-text-ellipsis'
 import Alphabet from './Alphabet'
 import Loading from '../Messages/Loading'
 import { getLinkObject } from '../../shared/ContentfulLibs'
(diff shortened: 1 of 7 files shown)


src/components/DatabaseList/presenter.js
```diff
@@ -1,5 +1,5 @@
 // Presenter component for a Page content type from Contentful
-import React from 'react'
+import React, { Component } from 'react'
```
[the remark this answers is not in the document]
danthewolfe: removed

src/components/DatabaseList/style.css
```diff
@@ -9,3 +9,13 @@
     display: inline;
   }
 }
+
+.database-list {
+  width: 100%;
+  word-wrap: break-word;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
hbeachey: I think the reason we didn't do this before was because the ellipsis would go past the text at the end of the column and look strange. Does something in this css cause that to not happen?

hbeachey: 👍

PR #464, opened 2018-04-25
Allow multiple alerts on pages

hbeachey:
Includes content types
- Pages
- Dynamic Pages
- Column Containers

Also altered the display of alerts on these pages so formatted better.

### DIFF
--- a/src/components/Contentful/Page/columnPresenter.js
+++ b/src/components/Contentful/Page/columnPresenter.js
@@ -57,7 +57,11 @@ const ColumnContainerPresenter = (props) => {
         description={page.shortDescription}
         image={false}
       />
-      <PageAlert alert={page.alert} />
+      <div className='columnAlertContainer'>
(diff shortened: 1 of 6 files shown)


src/components/Contentful/Page/columnPresenter.js
```diff
@@ -57,7 +57,11 @@ const ColumnContainerPresenter = (props) => {
         description={page.shortDescription}
         image={false}
       />
-      <PageAlert alert={page.alert} />
+      <div className='columnAlertContainer'>
+        <div className='col-md-12 col-xs-12'>
```
jonhartzler: Can we reverse this so that we could put the div inside of the component?

hbeachey: No, because this is only used on the "column" pages like /research
There are 3 different display cases for the page alerts. This is the most unfortunate one, so if there's a better way I'm totally open to changes.

PR #467, opened 2018-04-30
ErrorBoundary (LW-276 continued)

danthewolfe:
* Create an `ErrorBounday` component.
* Create higher order `withErrorBoundary` component.
* Apply higher order comopnent to a few problematic components that don't render with phantomJS.
* Unit tests.

**Usage**

* There are a couple of ways to use `ErrorBoundary`
* Use as a regular component:
```
<ErrorBoundary>
  <PossiblyBrokenComponent/>
</ErrorBoundary>
```

* Use as a [higher-order component](https://reactjs.org/docs/higher-order-components.html):
```
...
import {withErrorBoundary} from '<PATH>/ErrorBoundary'

const MyComponent = (props) => {

...

}
export default withErrorBoundary(MyComponent)
```

* Use as a higher-order component and pass a backup component to render in event of failure:
```
...
import {withErrorBoundary} from '<PATH>/ErrorBoundary'

const MyComponent = (props) => {

...

}

const AltComponent = () => { return <div>Something went wrong.</div> }

export default withErrorBoundary(MyComponent, AltComponent)
```

**Notes on testing**
* I've tried to make `withErrorBoundary` as transparent as possible. One change that may need to be is make is when creating the `enzymeWrapper` with `shallow`. You may need to make use of  [`dive()` ](https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/dive.md)with a depth of `1` to get the expected result. :

```
let enzymeWrapper = shallow(<TestComponent props={...props}/>)
```

to 

```
let enzymeWrapper = shallow(<TestComponent props={...props}/>).dive(1)
```


src/tests/components/ErrorBoundary/index.test.js
```diff
@@ -0,0 +1,76 @@
...
+  return <div>{props.testProp}</div>
+}
+
+const AlternativeComponent = () => {
+  return <div>Alternative</div>
+}
+
+const BrokenComponent = () => {
+  // do not return anything and throw error
+  throw new Error('I AM ERROR')
```
danthewolfe: i_am_error

danthewolfe:
@rfox2 Here is an example of how Error handling helps us catch ourselves with PhantomJS and potentially 'Fetch as Google'.

PhantomJS rendering of homepage capped at 1 second, no error handling:
capture_1525106796688

PhantomJS rendering of homepage capped at 5 seconds, no error handling (page is broken because of error fetching hours api):
capture_1525100040811

PhantomJS rendering of homepage capped at 5 seconds, with error handling (hours api error is caught by ErrorBoundary):
capture_1525099565039

PR #486, opened 2018-05-31
LW-454 Change search box on dev-staging to point to onesearchpprd

danthewolfe:
Point to onesearchpprd instead of onesearch whenever not on the production site.


danthewolfe:
Ah, good to know. I had expected there to be a JIRA ticket if that was the case. Adding one now and I will lump that work in with this pull request.

PR #487, opened 2018-06-04
LW-463 Reformat Advanced Search for NUI

danthewolfe:
* Make changes to search query based on description provided by ExLibris and expand to make it compatible with observed behavior of advanced search in NUI.
* https://knowledge.exlibrisgroup.com/Primo/Product_Documentation/050New_Primo_User_Interface/New_UI_Customization_-_Best_Practices#Creating_a_Search_Box_With_Deep_Links_to_the_New_UI

### DIFF
--- a/src/components/SearchDrawer/searchQueryBuilder.js
+++ b/src/components/SearchDrawer/searchQueryBuilder.js
@@ -1,18 +1,39 @@
 import { ONESEARCH, NDCATALOG, CURATEND, LIBRARY } from './searchOptions'
+
+const padLeftZero = (num) => {
+  if (num.length < 2) {
+    return `0${num}`
+  }
+  return num
+}
+
 const onesearchUrl = (queryTerm, isAdvanced, isOnesearch) => {
   const tab = isOnesearch ? 'onesearch' : 'nd_campus'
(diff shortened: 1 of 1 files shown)


src/components/SearchDrawer/searchQueryBuilder.js
```diff
@@ -43,51 +64,29 @@ const searchQuery = (searchStore, advancedSearch, history) => {
...
-    `&vl%281UIStartWith1%29=${precision1}` +
-    `&vl%28freeText1%29=${freeText1}` +
-    `&vl%28boolOperator1%29=${bool1}` +
-    `&vl%2816833819UI2%29=${scope2}` +
-    `&vl%281UIStartWith2%29=${precision2}` +
-    `&vl%28freeText2%29=${freeText2}` +
-    `&vl%2816772486UI3%29=${materialType}` +
-    `&vl%2824400451UI4%29=${language}`
+    // Build advanced search query
+    searchTerm = `query=${scope0},${precision0},${freeText0}`
```
hbeachey: Is there a problem with keeping the `boolX` at the end all the time (I think that's what NUI does)? That way we don't hit problems if someone skips one row.

danthewolfe: Moved bools to end of lines.

PR #494, opened 2018-06-27
Fixed print styles for floor map pages

awetherill:


jonhartzler:
As a note the travis failure is the timeout error.  Usually these fix themselves on the second run and the tests pass for me locally.

PR #518, opened 2018-08-29
LW-509 - added dropdown arrow in search scope dropdown

rdoughty:


rdoughty:
@jonhartzler - now validating prop types; hound resolved

PR #530, opened 2018-09-28
Fix preview links

jonhartzler:


danthewolfe:
Fix the two hound errors by changing `const url` to `let url`, the 👍

PR #534, opened 2018-10-03
Remove the info things we don't want from info.

jonhartzler:


src/components/SiteInfo/index.js
```diff
@@ -1,6 +1,4 @@
 import React, { Component } from 'react'
-import findIP from './findIP'
-import lastModified from './lastModified'
```
danthewolfe: Can you remove these two files?

PR #552, opened 2018-11-01
Show service point on floor map page according to the building it is in.

mlkirkpatrick:
If the ServicePoint component receives a reference with an id but is missing the full details (fields), go and fetch the service point explicitly from Contentful.


mlkirkpatrick:
Fixed the formatting in c1602ac. Can this be approved now?

PR #556, opened 2018-12-03
LW-461 Expired Aleph Account Notice

mlkirkpatrick:
Adds a message on the item requests page alerting patrons if the user's aleph account has expired. (Since aleph does not return an intelligent error if you try to renew and are not allowed to.)

### DIFF
--- a/src/tests/components/LoanResources/presenter.test.js
+++ b/src/tests/components/LoanResources/presenter.test.js
@@ -20,7 +20,7 @@ let props = {
 }
 describe('components/LoanResources/presenter.js', () => {
   beforeEach(() => {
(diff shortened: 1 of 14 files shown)


src/tests/components/LoanResources/presenter.test.js
```diff
@@ -20,7 +20,7 @@ let props = {
 }
 describe('components/LoanResources/presenter.js', () => {
```
danthewolfe: Channeling @jonhartzler here, but we should probably add a test for userLoading={true}. Not that we're doing a lot of testing in Usurper, but we probably should be.

PR #559, opened 2018-12-14
Fix response validation coming from contentful direct.

mlkirkpatrick:
We are supposed to be catching errors when fetching from contentful direct and piping the result into an error object, which has a different schema compared to the success object. However, we were frequently only checking for the existence of a response to determine whether there was a success or not, so errors were incorrectly getting marked as "statuses.SUCCESS" and subsequently causing rendering errors since the schema did not match.

### DIFF
--- a/src/actions/contentful/allAlerts.js
+++ b/src/actions/contentful/allAlerts.js
@@ -26,7 +26,7 @@ const receiveAllAlerts = (response) => {
   }
 
   try {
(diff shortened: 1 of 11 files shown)


danthewolfe:
@rfox2 For these particular cases. The main issue is we are trying to iterate over something that is expected to be an array. The hope is that this will stop the Sentry reports about `e.map is not a function` and similar such messages.

PR #565, opened 2019-01-22
LW-572 Add message to courses page.

mlkirkpatrick:
Added component to display alerts dynamically on non-contentful pages.


src/components/Courses/presenter.js
```diff
@@ -172,11 +172,17 @@ class Courses extends Component {
           <StaticBody slug='courses' preview={this.props.preview} />
 
           <div className='row'>
-            <div className='col-md-8 col-sm-7'>
+            <div className='col-md-8 col-sm-7' style={{ position: 'relative' }}>
```
danthewolfe: We should probably try to get out of the practice of setting style this way. I'd prefer to see a className here.

PR #568, opened 2019-01-23
LW-570 Refactor expired account message.

mlkirkpatrick:
Hopefully no more false-positives, and a better error message if the user's aleph account fails to load.

Includes a "PageAlert" component which can be plugged in to any page. This differs from the existing alerts in that those only work with Contentful pages or at the global level. This allows us to display or hide alerts dynamically on a particular page, but uses the same presenter as the Contentful alerts to maintain UI consistency.


src/components/Messages/PageAlert/index.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import Presenter from '../../Contentful/Alert/presenter.js'
+import { alertMap, alertCatagorize } from '../../Contentful/Alert/alertHelpers.js'
+
+class PageAlert extends React.Component {
```
jonhartzler: This is a nice class.  Are there other places it should be in use @danthewolfe ?

danthewolfe: Maybe, could we modify it to take a couple of props and use it for other alerts on the site?

mlkirkpatrick: I'm not sure what else one might want to add. It can already accept pretty much anything as children, and takes a type prop to switch between an "informational" or "warning" message, just like the Contentful alerts. There could be some additional types that are styled differently, but those should be easy enough to add to the class if we find a need to add more styles.

PR #592, opened 2019-04-10
Fix code smells.

mlkirkpatrick:
Fixes tons of code quality issues as reported by code climate and linter.


src/components/LoanResources/ResourceList/presenter.js
```diff
@@ -7,6 +7,10 @@ import ExportButton from './ExportButton'
 import DeleteButton from './DeleteButton'
 
 const ResourceList = (props) => {
+  const renewal = props.renewal
```
danthewolfe: Use spread operator instead and add `sortClass` and `listType`?

`const { renewal, alephId, deleteFromHistory, sortClick, sortClass, listType } = props`

.eslintrc
```diff
@@ -1,32 +1,45 @@
...
+  "extends" : [
     "standard",
     "standard-react",
     "react-app"
   ],
-  "plugins": [
+  "plugins" : [
     "babel",
     "react",
+    "react-hooks",
```
danthewolfe: You'll need to add `yarn add -D eslint-react-hooks`

PR #594, opened 2019-04-18
LW-621 Refactor Courses components.

mlkirkpatrick:
- Split out into subcomponents.
- Move CSS styles to a new file next to the component.
- Implement CSS modules.
- Expand/improve unit tests.

### DIFF
--- a/config/jest/fileTransform.js
+++ b/config/jest/fileTransform.js
@@ -6,6 +6,24 @@ const path = require('path')
 
 module.exports = {
   process (src, filename) {
+    const assetFilename = JSON.stringify(path.basename(filename))
+
+    if (filename.match(/\.svg$/)) {
+      return `module.exports = {
+        __esModule: true,
+        default: ${assetFilename},
+        ReactComponent: (props) => ({
+          $$typeof: Symbol.for('react.element'),
+          type: 'svg',
+          ref: null,
+          key: null,
+          props: Object.assign({}, props, {
+            children: ${assetFilename}
+          })
+        }),
+      };`
+    }
+
     return 'module.exports = ' + JSON.stringify(path.basename(filename)) + ';'
   },
 }
--- a/src/components/Account/Courses/CourseList/CourseCard/index.js
+++ b/src/components/Account/Courses/CourseList/CourseCard/index.js
@@ -1,0 +1,70 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import t from 'typy'
+import Link from 'components/Interactive/Link'
+import styles from '../style.module.css'
+
+const makePathFinders = (course) => {
+  const arr = t(course, 'pathfinders').isArray ? course.pathfinders : (course.pathfinder ? [ course.pathfinder ] : [])
+  if (!arr.length) {
+    return null
+  }
+
+  return (
+    <ul>
+      { arr.map((item) =>
+        <li key={item.url}>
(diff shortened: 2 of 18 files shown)


src/components/Account/Courses/CourseList/CourseCard/index.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import t from 'typy'
+import Link from 'components/Interactive/Link'
+import styles from '../style.module.css'
+
+const makePathFinders = (course) => {
```
danthewolfe: Move to its own file and name it PathFinders.

src/components/Account/Courses/CourseList/CourseCard/index.js
```diff
@@ -0,0 +1,70 @@
...
+    <div className={styles.courseCard}>
+      <div className={styles.courseColumn}>
+        <p className={styles.courseName}>{props.course.title}</p>
+        <small className={styles.courseSubtitle}>{subtitle}</small>
+      </div>
+      <div className={styles.courseReserves}>
+        {courseReserves}
+      </div>
+      <div className={styles.courseResources}>
+        {pathfinders}
```
danthewolfe: <PathFinders course={props.course} />

src/components/Account/Courses/index.js
```diff
@@ -1,32 +1,25 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
-import getCourses from '../../../actions/personal/courses'
+import { bindActionCreators } from 'redux'
+import t from 'typy'
```
danthewolfe: As a convention we've been trying to use `import typy from 'typy'` since `t` is not very descriptive. Since it's the default export from the package, you can get away with that.

src/components/Account/Courses/presenter.js
```diff
@@ -1,147 +1,39 @@
...
-
-    if (out.length === 0) {
-      return (
-        <p className='noClasses'>
-          No classes to display for the current semester.
-        </p>)
-    }
-
-    return out
+    return (
```
danthewolfe: This is so much better to look at.

config/jest/fileTransform.js
```diff
@@ -6,6 +6,24 @@ const path = require('path')
 
 module.exports = {
   process (src, filename) {
+    const assetFilename = JSON.stringify(path.basename(filename))
+
+    if (filename.match(/\.svg$/)) {
+      return `module.exports = {
```
rfox2: Is the backtick intentional?

config/jest/fileTransform.js
```diff
@@ -6,6 +6,24 @@ const path = require('path')
...
+        ReactComponent: (props) => ({
+          $$typeof: Symbol.for('react.element'),
+          type: 'svg',
+          ref: null,
+          key: null,
+          props: Object.assign({}, props, {
+            children: ${assetFilename}
+          })
+        }),
+      };`
```
rfox2: I can see it's enclosed here - just wondering about the backtick vs. single or double quotes.

config/jest/fileTransform.js
mlkirkpatrick: Yes. Not using double-quotes because we've generally been avoiding that with javascript. Not using single quotes because there are several single quotes inside the text block and they would have to be escaped.

PR #608, opened 2019-05-24
Fix intermittent page rendering problem on secure pages

danthewolfe:
Reported problem: Secure pages intermittently get stuck on Loading animation.

Technical problem: `mapStateToProps` was not updating `this.props.login` correctly because of some internal optimizations that redux does durring mapStateToProps.

Solution: `this.props.login` is assigned as an immutable object with `Object.assign` instead of directly setting it to `state.personal.login`

https://redux.js.org/faq/react-redux#why-isn-t-my-component-re-rendering-or-my-mapstatetoprops-running


danthewolfe:
@mlkirkpatrick Let's work to get this deployed after the holiday weekend and discuss this issue. A similar situation may be lurking in a few other places.

PR #624, opened 2019-06-12
LW-683 Merge Settings into Favorites page

mlkirkpatrick:


package.json
```diff
@@ -115,6 +115,7 @@
         "hoursAPIURL": "http://hoursAPIURL.edu",
         "contentfulAPI": "http://contentfulAPI.edu",
         "userPrefsAPI": "http://userPrefsAPI.edu",
+        "serviceNowBaseURL": "http://serviceNow.com/test?id=123",
```
danthewolfe: Is this really supposed to have `/test?id=123`?

src/components/Account/CirculationHistory/CircHistorySidebar/index.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  return (
+    <aside className={'col-md-4 col-sm-5 col-xs-12 right ' + styles.checkoutHistorySidebar}>
+      <div>
+        <button className={'button callout ' + styles.optOutBtn} onClick={props.onClickOptOut}>
+          Opt-Out and Delete History
+        </button>
+        <PageLink className='button callout' cfPage={calloutLink} />
+        <h3 id='note'>NOTE:</h3>
```
danthewolfe: Not sure if "NOTE:" is the best heading. Maybe "Notice:" or "Please note:" Question for @lajamie or @mlharden  - can wait on QA for feedback.

src/components/Account/Favorites/index.js
```diff
@@ -30,8 +53,22 @@ class FavoritesContainer extends Component {
     } else if (this.props.login.redirectUrl) {
       window.location.replace(this.props.login.redirectUrl)
     }
-    if (this.props.loggedIn && this.props.favoritesStatus === statuses.NOT_FETCHED) {
-      this.props.getAllFavorites()
+    if (this.props.loggedIn) {
```
danthewolfe: This might be a good place to practice variable extraction:

```
const cases = [
  { statusType: 'favoritesStatus', action: 'getAllFavorites` },
  ... // rest of the cases go here
]
cases.forEach(case => {
  if(this.props[case.statusType] === statuses.NOT_FETCHED) {
    this.props[case.action]()
  }
})
```

You could also go a step further with method extraction and add:

```
checkAndClear(statusType, action) {
  if(this.props[statusType] === statuses.NOT_FETCHED) {
    this.props[action]()
  }
}
```
which would make the loop look like this:
```
const cases = [
  { statusType: 'favoritesStatus', action: 'getAllFavorites` },
  ...
]
cases.forEach(case => {
  this.checkAndClear(case.statusType, case.action)
})
```

Ditto for the set of if checks above for the `clearUpdateSettings`.  Having state might make it a bit more complicated.

src/components/Account/Favorites/presenter.js
```diff
@@ -13,9 +16,19 @@ const Presenter = (props) => {
   const dbItems = (props.dbFavorites && props.dbFavorites.items) ? props.dbFavorites.items : []
   const subjectItems = (props.subjectFavorites && props.subjectFavorites.items) ? props.subjectFavorites.items : []
   const loading = (props.favoritesStatus === statuses.NOT_FETCHED || props.favoritesStatus === statuses.FETCHING)
+  const sidebar = (
+    <SideNav className='side-nav-bg'>
+      <ul>
+        <a className='side-anchors' href={'#manage_' + KIND.databases}><li>Databases</li></a>
+        <a className='side-anchors' href={'#manage_' + KIND.subjects}><li>Subjects</li></a>
+        <a className='side-anchors' href={'#preferredLocation'}><li>Preferred Location</li></a>
+        <a className='side-anchors' href={'#homePageDisplay'}><li>Home Page Display</li></a>
```
danthewolfe: Maybe extract these to a subcomponent to reduce dupllication.

package.json
mlkirkpatrick: The "baseURL" that is actually used includes an id which identifies our service now instance. It could be argued that the implementation should be able to handle a url that doesn't already have a first query param on it, but this test does mirror our service now URLs in production use.

src/components/Account/CirculationHistory/CircHistorySidebar/index.js
mlkirkpatrick: ¯\_(ツ)_/¯ it's been on the page for weeks (months?) and nobody has suggested I change it.

src/components/Account/Favorites/index.js
mlkirkpatrick: Good idea. I do think that would make things a bit cleaner.

package.json
danthewolfe: 👍

src/components/Account/CirculationHistory/CircHistorySidebar/index.js
danthewolfe: 👍 Works for me.

PR #644, opened 2019-07-03
LW-706 Send Contentful preview requests to livequery to reduce cache time.

mlkirkpatrick:
Requests to Contentful Direct were still being cached regardless of whether they had the preview flag or not. Since previews could be cached, this meant Contentful users could not get quick feedback from the preview feature as intended.

Sending all previews to livequery will greatly reduce the cache time (currently 30 seconds), which should be a much better experience when trying to rapidly prototype changes in Contentful.

### DIFF
--- a/src/actions/contentful/entry.js
+++ b/src/actions/contentful/entry.js
@@ -54,7 +54,7 @@ export const fetchEntry = (id, slug, preview) => {
     identifierParam = encodeURIComponent(`sys.id=${id}&include=3`)
     entryIdent = id
   }
-  let url = `${Config.contentfulAPI}/query?locale=en-US&query=${identifierParam}`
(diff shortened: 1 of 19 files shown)


src/actions/contentful/entry.js
```diff
@@ -54,7 +54,7 @@ export const fetchEntry = (id, slug, preview) => {
     identifierParam = encodeURIComponent(`sys.id=${id}&include=3`)
     entryIdent = id
   }
-  let url = `${Config.contentfulAPI}/query?locale=en-US&query=${identifierParam}`
+  let url = `${Config.contentfulAPI}/${preview ? 'livequery' : 'query'}?locale=en-US&query=${identifierParam}`
```
danthewolfe: Can we extract this url pattern maybe so we can update it in a single place?

mlkirkpatrick: Good idea! Consider it done.

PR #647, opened 2019-07-09
LW-709 Add description for rearranging favorites.

mlkirkpatrick:


src/components/Account/Favorites/ManageFavorites/presenter.js
```diff
@@ -18,6 +21,11 @@ const ManageFavorites = (props) => {
       <div className='section-box pad-edges'>
         <div className='row'>
           {props.message}
+          <br /><br />
```
danthewolfe: Would prefer a css solution to double `<br/>`, but okay with this for now.

PR #658, opened 2019-07-16
Fixed display of the search box in Safari.

awetherill:


src/static/css/global.css
```diff
@@ -1313,6 +1313,7 @@ label[for='q']:after, .uSearchBox label:after {
   width: 100%;
   font-size: 18px;
   border: none;
+  -webkit-appearance: textarea;
```
mlkirkpatrick: I think this should be textfield instead of textarea.

Furthermore, this does not address search fields across the whole site. Anything with input[type='search'] is showing up incorrectly. (See Databases A-Z, Items & Requests, Checkout History, Favorites, etc.)

awetherill: Oh, good catch. Just pushed that change.

PR #670, opened 2019-08-07
Adjusted margins on H2s and changed the styling for the filter button…

awetherill:
…s on the Database A-Z pages.

### DIFF
--- a/src/static/css/global.css
+++ b/src/static/css/global.css
@@ -2455,6 +2455,10 @@ main h2, main h3, main h4, main h5 {
   line-height: 1.5em;
 }
 
+main div > h2:first-child {
+  margin-top: 0;
+}
+
 h3:first-child {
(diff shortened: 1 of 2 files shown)


src/static/css/global.css
```diff
@@ -3833,6 +3837,12 @@ img.favorite.white {
   width: 250px;
 }
 
+.style_dbSubject__2ESmF {
```
danthewolfe: This class name is generated from javascript and will likely change. It needs to be changed here:
`usurper/src/components/DatabaseList/style.module.css`
https://github.com/ndlib/usurper/blob/f7db40c50ab1210dabf3329fecc57b48e5559d46/src/components/DatabaseList/style.module.css#L20

awetherill: Updated

PR #686, opened 2019-09-04
Added spacing around the contact info box at the bottom of a news ite…

awetherill:
…m as well as adding some text styling to differentiate it.

### DIFF
--- a/src/components/Contentful/News/style.css
+++ b/src/components/Contentful/News/style.css
@@ -1,0 +1,13 @@
+.contactNews {
+	margin: 6em 0;
+    padding: 16px 10px 0;
+    border:  solid #aaa;
+    border-width: 2px 0;
+    color: #6F7275;
+}
+
+.contactNews p {
+	color: #6F7275;
+	font-size: 14px;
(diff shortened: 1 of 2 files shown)


src/components/Contentful/News/style.css
```diff
@@ -0,0 +1,13 @@
+.contactNews {
+	margin: 6em 0;
+    padding: 16px 10px 0;
+    border:  solid #aaa;
+    border-width: 2px 0;
+    color: #6F7275;
+}
+
+.contactNews p {
```
mlkirkpatrick: Please don't rely on the child being a p tag. We need to change to the LibMarkdown element to use a div or a span, but I haven't done it yet because of css that is specifically looking for a p tag. <p> tags are problematic because markdown will create child elements, many of which (such as a <div>) are not actually valid HTML inside a <p>. Browsers will render it anyway for historical reasons, but it triggers lots of warnings and should be remediated in the future.

awetherill: Updated.

PR #692, opened 2019-09-16
LW-756 Add Best Bets ("Featured Resources") to Databases list.

mlkirkpatrick:


danthewolfe:
Much wow! So tested! 🐕

PR #694, opened 2019-09-19
Adjusted spacing around contact info for News stories and made an adj…

awetherill:
…ustment to the print styles so floorplans print correctly on one sheet.


src/static/css/print.css
```diff
@@ -2,10 +2,12 @@
 	* {
 		font-family: sans-serif !important;
 	}
-	.top, #footer-links, #footer-info, .floor-info p, .floor-info h3, .floor .sp-address {
+	.top, #footer-links, #footer-info, .floor-info p, .floor-info h3, .floorplan .sp-address, .floorplan p, .floorplan h3, .floorplan .sp-address {
```
mlkirkpatrick: You've got `.floorplan .sp-address` in there twice, and I think the other two added are unnecessary. Also, I wouldn't think you'd want to hide h3 unilaterally in case something changed like the Map Key became an h3 and we likely would not catch the fact that it is hidden on print...

I also changed the class "sp-address" (since that was misleading and really encompasses the whole service point, not just the address line) in another PR, so there will be a merge conflict. But I can just as easily fix my commit by including your change with the new naming if this gets merged first. :)

awetherill: I *think* I've got this cleaned up now.

PR #699, opened 2019-10-01
LW-792 Include search results that Google filters by default.

mlkirkpatrick:
This also includes a usability improvement that will display the paging links for previous pages even when there are no results. This is helpful when users land on a page past the final result (due to inaccurate result counts), so they can try the page before it instead of returning to page 1 by running the search again.

### DIFF
--- a/src/actions/search.js
+++ b/src/actions/search.js
@@ -87,12 +87,7 @@ export const fetchResults = (queryString) => {
   return dispatch => {
     dispatch(siteSearchRequest(queryString))
     const qs = queryString ? QueryString.parse(queryString.replace('?', '')) : ''
-    fetch(
-      'https://www.googleapis.com/customsearch/v1?key=' + Config.gcseKey +
-      '&cx=' + Config.gcseCx +
-      '&q=' + qs.q +
-      '&start=' + (qs.start ? qs.start : '1')
-    )
+    fetch(`https://www.googleapis.com/customsearch/v1?key=${Config.gcseKey}&cx=${Config.gcseCx}&q=${qs.q}&start=${qs.start || '1'}&filter=0`)
       .then(response => response.json())
       .then(json => dispatch(siteSearchResponse(json)))
   }
--- a/src/actions/search.js
+++ b/src/actions/search.js
@@ -86,8 +86,19 @@ export const siteSearchResponse = (json) => {
 export const fetchResults = (queryString) => {
   return dispatch => {
     dispatch(siteSearchRequest(queryString))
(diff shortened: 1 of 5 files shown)


src/actions/search.js
```diff
@@ -87,12 +87,7 @@ export const fetchResults = (queryString) => {
   return dispatch => {
     dispatch(siteSearchRequest(queryString))
     const qs = queryString ? QueryString.parse(queryString.replace('?', '')) : ''
-    fetch(
-      'https://www.googleapis.com/customsearch/v1?key=' + Config.gcseKey +
-      '&cx=' + Config.gcseCx +
-      '&q=' + qs.q +
-      '&start=' + (qs.start ? qs.start : '1')
-    )
+    fetch(`https://www.googleapis.com/customsearch/v1?key=${Config.gcseKey}&cx=${Config.gcseCx}&q=${qs.q}&start=${qs.start || '1'}&filter=0`)
```
jonhartzler: This is a form of suggestion do with it as you will.  I might capture the variable with all the interpolation in one step and then use it in the next for readability and not collapsing of 2 disparate operations into one line.  Also i am a little unsure about the intepolation of a conditional statement inside a single line that adds extra burden to any one reading this line.

src/components/SearchPage/SearchPager/index.js
```diff
@@ -2,68 +2,74 @@ import React from 'react'
...
+  // This allows us to show previous page links if the user ends up on a page past the last record.
+  if (currentPage === 1 && !resultCount) {
+    return null
+  }
 
-    let startPage = 1
-    if (currentPage > 2) {
-      startPage = currentPage - 2
-    }
+  let lastPage = currentPage + 4
```
jonhartzler: Let's leave a comment for future us about why we are doing this checking.

src/components/SearchPage/presenter.js
```diff
@@ -23,9 +23,11 @@ class SearchPage extends Component {
   render () {
     let displayQuery = ''
     let pagerQuery = '/search?q='
+    let start = 1
     if (this.props.query) {
       displayQuery = this.props.query.replace(/(&start=\d+)/, '').replace('?q=', '')
       pagerQuery += displayQuery
+      start = parseInt(this.props.query.replace(/(\?q=.*(?=&))/, '').replace('&start=', '') || '1', 10)
```
jonhartzler: another readability suggestion.  Make a method out of dense lines that you can rename.  This allows for me to read the block here more naturally and understand what is going on and bunce to the method if I care about that algorithm.  as a bonus, it becomes easier to test just that regular expression logic 

such as 
start = determineStartPageFromQueryParams(..)

PR #729, opened 2019-10-28
Changed exclude to account for missing files

pweikel:


mlkirkpatrick:
LW-808

This excludes secure pages from the sitemap so they don't get indexed.

PR #739, opened 2019-11-15
LW-820 User Preferences refactor

mlkirkpatrick:
Reworked calls to user preferences to accommodate changes in data storage and retrieval. This PR coincides with ndlib/user_preferences#19.

### DIFF
--- a/config/configurationGen.js
+++ b/config/configurationGen.js
@@ -22,8 +22,8 @@ const genConfig = () => {
       favoritesEnabled: parameters.favoritesEnabled || false,
       subjectFilteringEnabled: parameters.subjectFilteringEnabled || false,
       eventsFilteringEnabled: parameters.eventsFilteringEnabled || false,
+      loginEnabled: parameters.loginEnabled || false,
     },
-    loginEnabled: parameters.loginEnabled,
   }
 
   const config = {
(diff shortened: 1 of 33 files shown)


danthewolfe:
Applied what I believe to be necessary changes as described above ^.

**NOTE:**
Delete of individual circ history items or delete of all circ history through the UI does not work. However, using Postman I experience the same behavior so I believe the issue is with user_preferences.

danthewolfe:
Add option to "disable" login. …
a176370
* Add the flag `--disableLogin` when running the `buildConfig` command: `node 
buildConfig.js stage=dev --disableLogin` OR add the line `loginEnabled: 
false`, to `configParameters.js`. 
* This will hide the "Log in"/"My Account" button and will display a 
message that accounts are not available on any account page if the user 
is already logged in.
* **NOTE** when running this branch for the first time after pulling, 
make sure to rerun `buildConfig` normally OR add `loginEnabled: true,` 
to your `configParameters.js` file.

mlkirkpatrick:
> **Add option to "disable" login.**
> 
> * Add the flag `--disableLogin` when running the `buildConfig` command: `node  buildConfig.js stage=dev --disableLogin` OR add the line `loginEnabled:  false`, to `configParameters.js`.

I have moved this flag to parameter store instead of putting it in the `buildConfig.js` script. This is how the other config flags work and it works better with the pipeline. It still functions and the same manner and will hide the My Account menu (as well as favorites on the home page) if `loginEnabled` is set to false. As such, **be sure to run `buildConfig.js` or add the value to `configParameters.js` after this is merged,** as Dan noted above.

PR #758, opened 2019-12-17
LW-828 Tweak styling of recurring indicator on event page.

mlkirkpatrick:


mlkirkpatrick:
Included in #761 so this PR is no longer needed.

PR #784, opened 2020-02-04
LW-870 Catch exceptions if local storage is unavailable.

mlkirkpatrick:


src/actions/personal/token.js
```diff
@@ -18,7 +18,11 @@ const oktaConfig = {
 export const initLogin = () => {
   return () => {
     // Save the user's requested url to local storage so we can redirect them after authenticating
-    window.localStorage.setItem('redirectUrl', window.location.href)
+    try {
+      window.localStorage.setItem('redirectUrl', window.location.href)
```
danthewolfe: This should technically probably be placed in `sessionStorage` instead of `localStorage`

mlkirkpatrick: You're right. I'll change it for posterity.

PR #790, opened 2020-02-07
LW-881 - Replaced heading tag with div containers in Hours component

pweikel:
This should realign the header hierarchy in several areas across the website


src/components/Hours/Current/presenter.js
```diff
@@ -32,27 +32,27 @@ const Presenter = ({ hoursEntry, openStatus, expanded, toggleExpanded, children
   return (
     <section className={`service-point ${statusClassName(openStatus)}`} itemScope itemType={schemaType(hoursEntry.servicePoint.type)} role='tablist'>
       <button className={`custom-style ${expanded ? 'collapse' : 'expand'}`} tabIndex={0} onClick={toggleExpanded} onKeyDown={toggleExpanded}>
-        <h4 aria-label={`${hoursEntry.name}, Open ${todayLabel}`} className='sp'>
+        <h2 aria-label={`${hoursEntry.name}, Open ${todayLabel}`} className='sp'>
           <div className='location' itemProp='name'><h2>{hoursEntry.name}</h2></div>
```
danthewolfe: Line 36, should probably not have this second set of `<h2 />` tags around `{hoursEntry.name}` either.

pweikel: To satisfy the requirement for WCAG 2.0 the following must be true:

- _The header following an h2 is h1, h2 or h3._
- _The header following an h1 is h1 or h2._

Currently, the header in question is an H4. If it were a H2 it should satisfy the requirements in both situations highlighted above. Is there a place on the website where an H4 is needed in this component? If not, I would suggest making this update for now and then add the `headingLevel` prop at some later point, if we still think we need it.

pweikel:
The usage of a header tag in this component deviates from the semantic purpose of heading tags which is to introduce a topic followed by related content, typically in a paragraph. In all but one of the 63 instances where this component appears an h2 is already present that defines what the current hours are referencing. This looks more like a container than a heading.

Hesburgh Library 1st Floor
Ask Us Desk
DVDs
Radiation Lab Reading Room

src/components/Hours/Current/presenter.js
```diff
@@ -32,27 +32,27 @@ const Presenter = ({ hoursEntry, openStatus, expanded, toggleExpanded, children
   return (
     <section className={`service-point ${statusClassName(openStatus)}`} itemScope itemType={schemaType(hoursEntry.servicePoint.type)} role='tablist'>
       <button className={`custom-style ${expanded ? 'collapse' : 'expand'}`} tabIndex={0} onClick={toggleExpanded} onKeyDown={toggleExpanded}>
-        <h4 aria-label={`${hoursEntry.name}, Open ${todayLabel}`} className='sp'>
-          <div className='location' itemProp='name'><h2>{hoursEntry.name}</h2></div>
+        <div aria-label={`${hoursEntry.name}, Open ${todayLabel}`} className='sp'>
+          <div className='location' itemProp='name'><div>{hoursEntry.name}</div></div>
```
danthewolfe: This inner `<div/>` should be able to be removed completely:
```
<div className='location' itemProp='name'>{hoursEntry.name}</div>
```

pweikel:
My bad! Thought I did that! Should be good now.

PR #797, opened 2020-02-12
Table component w/ Floor list data

pweikel:
Not yet connected to Contentful. Looking for feedback on the component and direction on the general setup within Usurper.


src/components/Contentful/Floorlist/common/tableBody.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+  render () {
+    const { data, columns } = this.props
+
+    return (
+      <tbody>
+        {data.map(item => (
+          <tr key={item._id}>
+            {columns.map(column =>
+              <td role='cell' key={this.createKey(item, column)}>{this.renderCell(item, column)}</td>)}</tr>
```
mlkirkpatrick: Closing parentheses/braces and tags should usually be on their own line so their indention is aligned with where they open. This makes it easier to read.

src/components/Contentful/Floorlist/common/tableBody.jsx
```diff
@@ -0,0 +1,36 @@
...
+    )
+  }
+}
+
+export default TableBody
+
+TableBody.propTypes = {
+  columns: PropTypes.array.isRequired,
+  data: PropTypes.array.isRequired,
+}
```
mlkirkpatrick: `PropTypes.array` satisfies the requirement for prop validation, but it's a bit vague. React can validate the contents of props passed into the component, but it can only do as much as you tell it. For instance, if you know the array is a list of strings, you can tell it so and React will be able to verify that, giving you relevant warnings if some unexpected data is passed in.
`example: PropTypes.arrayOf(PropTypes.string).isRequired,`
Also, you may have a nested object, but you know that certain properties are required. You can tell React the exact shape in as much detail as the component needs to know:
```
example: PropTypes.arrayOf(PropTypes.shape({
  field1: PropTypes.string.isRequired,
  field2: PropTypes.shape({
    anotherProp: PropTypes.number,
  }).isRequired,
})
```

In some cases, the "shape" of the object or array may not be relevant to the component that is using it, and instead is just being passed down to a child component. In that case, it is perfectly fine (and preferred, in my opinion) to just use `PropTypes.array` or `PropTypes.object`. (For instance, the Table component does not need to know the shape of the columns array because it doesn't do anything with it, but the `TableBody` expects those objects to have a `path` property.)

src/components/Contentful/Floorlist/common/tableHeader.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+
+class TableHeader extends Component {
+  render () {
+    return (
+      <thead>
+        <tr className='clickable'>
+          {this.props.columns.map(column => (
+            <th scope='col' role='heading' aria-describedby='caption' key={column.path || column.key}>{column.label}</th>))}
```
mlkirkpatrick: `aria-describeby` value needs to be an `id`, and "caption" is too broad to be unique, so I don't think this will work quite right.

src/components/Contentful/Floorlist/floorsTable.jsx
```diff
@@ -0,0 +1,34 @@
...
+import Table from './common/table'
+import { Link } from 'react-router-dom'
+import PropTypes from 'prop-types'
+
+class FloorsTable extends Component {
+  columns = [
+    {
+      path: 'name',
+      label: 'Floors',
+      content: floor => <Link to={`/floor/${floor.slug}`}>{floor.name}</Link>,
```
mlkirkpatrick: I might be missing something, but this `content` property seems redundant since you are passing the `data` property for the cell contents.

src/components/Contentful/Floorlist/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+    floors: [],
+    genres: [],
+    caption: '',
+  };
+
+  componentDidMount () {
+    const headings = [{ _id: '' }, ...getHeadings()]
+
+    this.setState({ floors: getFloors(), headings, caption: getCaption() })
+  }
```
mlkirkpatrick: You don't need state here; you could just call the functions inside the render function. In case it is not already clear, you won't need state when calling Contentful either. You'll need to do an API call on mount, but the results will go the store which can then be mapped to props. So the whole render happens from prop values.

Maybe you were aware of that and you were just trying to roughly replicate the store functionality, but I thought it would point it out in case you were not aware. We rarely use state. It's mostly for interactive elements that are going to change while the user is on a page, rather than for storing read only data.

src/components/Contentful/Floorlist/common/table.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import TableHeader from './tableHeader'
+import TableBody from './tableBody'
+import PropTypes from 'prop-types'
+
+const Table = ({ columns, data, caption }) => {
+  return (
+    <table role='table' >
+      <caption id={caption} />
```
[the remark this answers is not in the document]
pweikel: I didn't want it to display publicly. It's really there for the benefit of users using screen readers. Putting it inside the tag displays it publicly.

package.json
```diff
@@ -34,6 +34,7 @@
     "redux-mock-store": "^1.5.3",
     "redux-thunk": "^2.3.0",
     "reselect": "^4.0.0",
+    "serialize-javascript": "^2.1.1",
```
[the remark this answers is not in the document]
pweikel: I don't know. When I published my branch, GitHub sent me this security notice, so I upgraded like it suggested. I'm not using it. I don't know what it is. 
Screen Shot 2020-02-12 at 4 55 52 PM

src/components/App/index.js
```diff
@@ -63,7 +64,7 @@ const store = createStore(
 const App = (props) => {
   return (
     <Switch>
-      { Rewrite(props) }
+      {Rewrite(props)}
```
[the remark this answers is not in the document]
pweikel: I'll look into a similar plugin. That would be most helpful!

src/components/Contentful/Floorlist/common/style.css
```diff
@@ -0,0 +1,71 @@
...
+  justify-content: center;
+  font-size: calc(10px + 2vmin);
+  color: white;
+}
+
+.App-link {
+  color: #09d3ac;
+}
+
+table {
```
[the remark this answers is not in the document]
pweikel: I've a different branch where I have updated the Global CSS file to include new table styles.  I haven't seen any unwieldy behavior it yet. I'll create a pull request for it. I would prefer to just add this to the Global css file seeing as how table is supposed to be universal. 

CSS modules sound interesting. It isn't necessary here but I'll look into it for future component design.

src/components/App/index.js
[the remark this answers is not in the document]
pweikel: Prettier is disabled and I'm using ESLint 2.0.15 within Usurper. It seems to like to correct the extra space here.

src/components/Contentful/Floorlist/common/table.jsx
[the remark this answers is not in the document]
pweikel: This isn't a recommended usage of the caption element. I changed it to display publicly.

package.json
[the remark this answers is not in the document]
mlkirkpatrick: Ok, but since it wasn't already in package.json that means it was a dependency that was referencing an outdated version of serialize-javascript. Adding it to the package isn't really going to help, because you will just end up with multiple versions of the node package, and the dependency will continue to use the old version.

In any case, you should be able to remove the explicit dependency from our package. The dependency is related to webpack, which can be a bit of a chore to upgrade, so I suggest saving that for a separate ticket.

src/components/App/index.js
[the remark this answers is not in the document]
mlkirkpatrick: It may not be reading our `.eslintrc`, or maybe the VSCode plugin behaves differently than the Atom one. May take some fiddling for us to figure out what the difference is.

src/components/Contentful/Floorlist/common/tableBody.jsx
```diff
@@ -0,0 +1,36 @@
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
+import _ from 'lodash'
+
+class TableBody extends Component {
```
danthewolfe: Unless there is a strong need to include life cycle methods, our preference is to declare react component as functions.

```
const TableBody = (props) => {
  ...
}
```

src/components/Contentful/Floorlist/common/tableBody.jsx
pweikel: Do you think it would be better to leave it vague? The content that could be regurgitated through the table component may come in from varied sources and it may not always be strings inside of an array. What do you think?

src/components/Contentful/Floorlist/common/tableHeader.jsx
pweikel: Right. This should probably reference the id of the tableBody.

src/components/Contentful/Floorlist/common/tableBody.jsx
mlkirkpatrick: It depends how the component is written. If it can accommodate any kind of data type, sure. In your current code, the `data` prop is expected to be an array of objects that **must** contain an `id` property, otherwise you will not get valid keys. So at minimum, I think it should specify that:
```
data: PropTypes.arrayOf(PropTypes.shape({
  id: PropTypes.string.isRequired,
})).isRequired,
```
Or to allow for an id that is a number (which can be implicitly cast to a string):
```
data: PropTypes.arrayOf(PropTypes.shape({
  id: PropTypes.oneOfType([
    PropTypes.string,
    PropTypes.number,
  ]).isRequired,
})).isRequired,
```

src/components/Contentful/Floorlist/common/table.jsx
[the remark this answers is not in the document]
danthewolfe: Also, a little note: the `id` attribute should always be unique within a page or you will get accessibility errors. Since we try to write reusable components whenever possible, it's easy to accidentally pass in a non-unique value through props, especially if you end up iterating through a list and rendering multiple versions of a component, so I generally recommend avoiding setting an `id` through props.

src/components/Contentful/Floorlist/common/tableBody.jsx
```diff
@@ -0,0 +1,36 @@
...
+  renderCell = (item, column) => {
+    if (column.content) return column.content(item)
+
+    return _.get(item, column.path)
+  };
+
+  createKey = (item, column) => {
+    return item._id + (column.path || column.key)
```
danthewolfe: This is mostly a preference, but I try to avoid string concatenation with the `+` operator. Using back tic quotes:
```
return `${item._id}${column.path || column.key}`
```
This is not a huge issue since you're just putting together two variable, but when you have something longer I think it's easier to read, or at least mentally parse. Example:
```
const url = some.baseUrl + '/' + some.path.parameter + '/' + 'som.path.id' + '/edit'
```
vs:
```
const url = `${some.baseUrl}/${some.path.parameter}/${some.path.id}/edit`
```

src/components/Contentful/Floorlist/common/tableBody.jsx
danthewolfe: https://reactjs.org/docs/typechecking-with-proptypes.html

src/components/Contentful/Floorlist/common/tableBody.jsx
pweikel: Got it. Makes sense!

danthewolfe:
> File names and folder structure should follow the conventions of the rest of the project.
> 
> * Components should be named in [pascal case](http://wiki.c2.com/?PascalCase) (E.g. FloorList instead of Floorlist.)
> * Components are a directory with an `index.js` file (and optionally a `presenter.js` and/or `style.css`.) On rare occasions there may be supplementary files as well that are relevant only to that component, but the most typical case is that these are child components and should thus be their own directory under the parent. (Examples: TableHeader, TableBody)

Usurper is a little weird in that it has a separate `testing` directory that mirrors the layout of the `src` directory. In most of our other React projects, unit tests live along side their component or other code that they test and are simply called `test.js`, which makes seeing what is tested and what isn't a lot easier.

Screen Shot 2020-02-13 at 11 43 25 AM

PR #798, opened 2020-02-13
LW-861 Add map note to service point.

mlkirkpatrick:


src/components/Contact/ServicePoint/index.js
```diff
@@ -94,6 +94,11 @@ const Contact = (props) => {
     // this is hidden in css on the contact page
     accessNote = <LibMarkdown className='access-note'>{ sp.accessNote }</LibMarkdown>
   }
+  let mapNote
+  if (sp.mapNote) {
+    // this is hidden in css except on map page
+    mapNote = <LibMarkdown className='map-note'>{ sp.mapNote }</LibMarkdown>
+  }
```
danthewolfe: Since `mapNote` isn't reassigned and it's relatively simple thoughts on putting it in a single line so we can use `const` instead of `let`?
```
// this is hidden in css except on map page
const mapNote = sp.mapNote ? <LibMarkdown className='map-note'>{ sp.mapNote }</LibMarkdown> : null
```

mlkirkpatrick: Of course. :)

PR #799, opened 2020-02-13
LW-855 LW-856 Exhibits

mlkirkpatrick:
This includes both the home page display and the landing page. There are expected to be more changes to the landing page, but there is also a feature flag that will keep everything hidden until it is ready anyway.

There is some significant refactoring of landing pages in here. The goal was to get as much of the functionality (facets, searching, date selection, etc.) into a reusable wrapper so we do not duplicate logic between events and exhibits. (As well as any future landing pages.) I had already started this effort with the Past and Current events pages reusing the same wrapper, but now it is more broad so that other landing pages can use it as well.


config/configurationGen.js
```diff
@@ -22,6 +22,7 @@ const genConfig = () => {
       subjectFilteringEnabled: parameters.subjectFilteringEnabled || false,
       eventsFilteringEnabled: parameters.eventsFilteringEnabled || false,
       loginEnabled: parameters.loginEnabled || false,
+      exhibitsEnabled: parameters.exhibitsEnabled || false,
```
danthewolfe: I understand most of these enabled/disabled feature settings have been added as part of a way to keep features dark while they are under development and awaiting PO approval to release, but do we have plans to remove any of these parameters post feature launch? I can image this list growing huge over time, and other than maybe wanting to disable login for various reasons, I don't think we'd want to turn off new features after they launch very often.

src/components/Home/Events/presenter.js
```diff
@@ -10,18 +10,21 @@ import './style.css'
...
-            />
+            <React.Fragment key={entry.id}>
+              <EventCard
+                entry={entry}
+                showDescription={false}
+                showImage={false}
+                showTags={false}
+              />
+              { index < entries.length - 1 && (
+                <hr className='card-divider' />
```
danthewolfe: Can we get rid of the index (I know you're a fan) and just return `<hr className='card-divider' />` after every card and then in css we can add:
```
.card-divider:last-of-type {
  display: none;
}
```

config/configurationGen.js
mlkirkpatrick: Yes, I agree. I do intend to remove them at some point, and I was actually just thinking it's about time to take out the favorites one. Subjects filtering is still stuck in a holding state, but events filter just went live so I can probably remove that as well.

src/components/Home/Events/presenter.js
mlkirkpatrick: 👍nice suggestion

PR #801, opened 2020-02-14
LW-857 Change featured services to icons instead of quick links.

mlkirkpatrick:


src/components/Home/index.js
```diff
@@ -33,35 +34,42 @@ class Home extends Component {
         />
         <PageTitle title='Hesburgh Libraries' hideInPage />
         <SearchProgramaticSet open />
+        { Config.features.exhibitsEnabled && (
```
mlkirkpatrick: POs want all of the home page changes to go live at the same time, so that means waiting for exhibits to be ready. That's why it uses that flag, and will continue to display quick links until it is flipped.

PR #814, opened 2020-03-03
Lw 858 - Refactored Table component

pweikel:


src/components/Contentful/FloorList/FloorsTable/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import Table from '../../../Table/'
+import PropTypes from 'prop-types'
+
+function FloorsTable (props) {
```
mlkirkpatrick: Prefer es6 style arrow functions.
```
const FloorsTable = (props) => {
```

src/components/Contentful/FloorList/index.js
```diff
@@ -0,0 +1,85 @@
...
+        {typy(floor, 'fields.title').safeArray.map(link => {
+          return <Link key={floor.sys.id} to={`/floor/${floor.fields.slug}`}>{link}</Link>
+        })}
+      </React.Fragment>
+    )
+    const spacesText = (
+      <React.Fragment>
+        {typy(floor, 'fields.spacesLinks').safeArray.map(link => {
+          const linkPath = link.fields.url || link.fields.slug
+          const lastElement = link
```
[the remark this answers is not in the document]
pweikel: Removed

src/components/Contentful/FloorList/index.js
```diff
@@ -0,0 +1,85 @@
...
+          const comma = floor.fields.spacesText || ((floor.fields.spacesLinks.length > 1) && (floor.fields.spacesLinks.indexOf(link) !== (floor.fields.spacesLinks.length - 1))) ? ', ' : ' '
+          return <React.Fragment><Link key={floor.fields.spacesLinks.indexOf(link)} to={`${linkPath}`}>{link.fields.title}</Link>{comma}</React.Fragment>
```
mlkirkpatrick: It's a little hard to read on one line like this. Would be nice to reformat it like this:
```
return (
  <React.Fragment>
    <Link key={floor.fields.spacesLinks.indexOf(link)} to={`${linkPath}`}>{link.fields.title}</Link>
    {comma}
  </React.Fragment>
)
```

In addition, I would probably change the no comma case to return `null` instead of a space (`' '`). Unnecessary whitespace is typically invisible, but can be annoying when you copy and paste.  ;)

src/components/Contentful/FloorList/presenter.js
```diff
@@ -0,0 +1,29 @@
+// Presenter component for a displaying Floors List table
+import React from 'react'
+import PropTypes from 'prop-types'
+import 'static/css/global.css'
```
mlkirkpatrick: Don't import `global.css`. It's already included in the site layout (a higher up component).

src/components/Table/tableBody.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
```
mlkirkpatrick: TableBody is a component, so it should be a directory with an `index.js`.

components/Table/tableBody.js –> components/Table/TableBody/index.js

src/components/Table/tableBody.js
```diff
@@ -0,0 +1,24 @@
...
+import PropTypes from 'prop-types'
+import typy from 'typy'
+
+function TableBody (props) {
```
mlkirkpatrick: Arrow function

src/components/Table/tableHeader.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
```
mlkirkpatrick: Same as TableBody for file naming structure.

src/components/Table/tableHeader.js
```diff
@@ -0,0 +1,19 @@
...
+import PropTypes from 'prop-types'
+
+function TableHeader (props) {
```
mlkirkpatrick: Arrow function

src/components/Table/tableBody.js
```diff
@@ -0,0 +1,24 @@
...
+  return (
+    <tbody>
+      {props.data.map(item => (
+        <tr key={item.sys.id || props.data.indexOf(item)}>
```
mlkirkpatrick: This isn't safe enough, because an item which does not have `sys.id` likely does not have `sys` at all, and thus will cause an exception rather than causing the or condition to trigger. Also, the map function passes index as a param so you don't necessarily need to get it with the indexOf function.
```
{props.data.map((item, index) => (
  <tr key={typy(item, 'sys.id').safeString || index}>
```

If you want `sys.id` to be optional, the component needs to be able to render without it. If you want it to be required, then using the index as a fallback has no purpose, and can just be removed.

src/components/Table/tableHeader.js
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <thead>
+      <tr className='clickable'>
```
mlkirkpatrick: I don't remember seeing this class in the css and the name doesn't seem applicable, so I don't think it should be here.

src/components/Contentful/FloorList/FloorsTable/index.js
pweikel: Done!

src/components/Contentful/FloorList/index.js
pweikel: Much clearer! Thanks!

src/components/Contentful/FloorList/presenter.js
pweikel: Done!

src/components/Table/tableBody.js
pweikel: Done!

src/components/Table/tableBody.js
pweikel: Done!

src/components/Table/tableHeader.js
pweikel: Done!

src/components/Table/tableHeader.js
pweikel: Done!

src/components/Table/tableHeader.js
pweikel: Removed!

src/components/Table/tableBody.js
pweikel: I think keeping it optional is best in the off-chance that we pass some data through the table that isn't sourced from Contentful.

So basically, what Typy is doing here is just checking to see if sys.id is somewhere in the data.map. If it finds it, it will use it. If it doesn't find it, it won't through an exception error, it will just proceed thru the condition. Another excellent use case for Typy!

src/components/Contentful/FloorList/index.js
```diff
@@ -0,0 +1,84 @@
...
+          return <Link key={floor.sys.id} to={`/floor/${floor.fields.slug}`}>{link}</Link>
+        })}
+      </React.Fragment>
+    )
+    const spacesText = (
+      <React.Fragment>
+        {typy(floor, 'fields.spacesLinks').safeArray.map(link => {
+          const linkPath = link.fields.url || link.fields.slug
+          const comma = floor.fields.spacesText || ((floor.fields.spacesLinks.length > 1) && (floor.fields.spacesLinks.indexOf(link) !== (floor.fields.spacesLinks.length - 1))) ? ', ' : ' '
+          return <React.Fragment><Link key={floor.fields.spacesLinks.indexOf(link)} to={`${linkPath}`}>{link.fields.title}</Link>{comma}</React.Fragment>
```
mlkirkpatrick: The references in `floor.fields.spacesLinks` are actually other Contentful objects. That means they have their own `sys.id`, which makes for a better key than getting their index in the array.

pweikel: Replace the key with the Contentful ID

src/components/Contentful/FloorList/FloorsTable/index.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import Table from '../../../Table/'
```
danthewolfe: Can we change this to: 
```
import Table from 'components/Table'
```

In general I recommend avoiding relative path imports when we can do an absolute one, unless it's something that's very close to the function doing the importing. There's a couple of reason to do absolute paths when possible, but basically, if we do any kind of refactoring that would involve moving this component up or down a level in the folder structure we then need to modify the import statements, or if we want to copy this code as a starting point for a new component that may or may not exist at a similar folder structure level, we would similarly have to modify the import.

src/components/Contentful/FloorList/index.js
```diff
@@ -0,0 +1,89 @@
+// Container component for array of Floors derived from Contentful
+import { connect } from 'react-redux'
+import { bindActionCreators } from 'redux'
+import React, { Component } from 'react'
+import PropTypes from 'prop-types'
```
danthewolfe: By convention we usually try to do `React` as the first import in the file and `PropTypes` as the second if they're needed. Then followed by any `node_modules` imports and our custom components and functions, with image and css imports being last. This makes it a bit easier to scan through the imported dependencies.

src/components/Contentful/FloorList/index.js
```diff
@@ -0,0 +1,89 @@
...
+  const pathSlug = locationArray[locationArray.length - 1]
+  const searchParams = new URLSearchParams(ownProps.location.search)
+  const preview = searchParams.get('preview') === 'true'
+  const filteredFloors = state.cfAllFloors.json.filter(floor => {
+    return typy(floor, 'fields.building.fields.slug').safeString === ownProps.slug && floor.fields.slug !== 'multifloor-building' && floor.fields.slug !== 'hesburgh-14th-floor' && floor.fields.slug !== 'hesburgh-3rd-floor'
+  })
+  const sortedFloors = filteredFloors.sort(function (a, b) {
+    return a.fields.floorNumber - b.fields.floorNumber
+  })
+  const floorData = sortedFloors.map(function (floor) {
```
danthewolfe: For consistency, can we change these to arrow functions?
```
const sortedFloors = filteredFloors.sort((a, b) => {

 ...

const floorData = sortedFloors.map((floor) => {
```

src/components/Contentful/FloorList/FloorsTable/index.js
pweikel: Good call. done!

src/components/Contentful/FloorList/index.js
pweikel: Done!

src/components/Contentful/FloorList/index.js
pweikel: Done!

PR #901, opened 2020-07-15
added fetch to pull directory data into the store for the contact in …

pweikel:
…meetingSpaceTable

Michael, thanks for all your help on this.


src/components/DynamicPages/RoomReservations/MeetingSpacesTable/index.js
```diff
@@ -49,17 +51,18 @@ const MeetingSpacesTable = (props) => {
     <Table columns={columns} data={tableData} className='meetingSpacesTable' />
   )
 }
-export default MeetingSpacesTable
+export default (MeetingSpacesTable)
```
danthewolfe, line 54: Not sure why `()` were added.

src/components/DynamicPages/RoomReservations/MeetingSpacesTable/index.js
```diff
@@ -49,17 +51,18 @@ const MeetingSpacesTable = (props) => {
...
       title: PropTypes.string.isRequired,
-      contact: PropTypes.object,
+      contact: PropTypes.string,
       photo: PropTypes.object,
       additionalFeatures: PropTypes.array,
       floor: PropTypes.object.isRequired,
       cardinalDirection: PropTypes.string.isRequired,
     }).isRequired,
   })).isRequired,
+  contactInfo: PropTypes.array.isRequired,
```
danthewolfe, line 67: Should this also be `arrayOf` with shape of an array of `contacts` with a `netid`?

src/components/DynamicPages/RoomReservations/index.js
```diff
@@ -32,22 +35,32 @@ export const mapStateToProps = (state, ownProps) => {
...
+
+    if (this.props.cfStaticStatus === statuses.NOT_FETCHED || this.props.cfStatic.slug !== this.props.pathSlug) {
       this.props.fetchSidebar(this.props.pathSlug, this.props.preview)
     }
+
+    if (this.props.contactInfo.status === statuses.NOT_FETCHED && this.props.allMeetingSpacesStatus === statuses.SUCCESS) {
+      const idData = this.props.meetingSpacesData.map(data => {
+        return data.fields.contact
+      })
+      const netids = [...new Set(idData)].filter(data => data)
```
danthewolfe, line 61: What is the advantage of using Set here? Is it just to ensure you have an empty array to work with if isData is null? If that's the case is the `filter` really necessary since it will return all items?

I don't recall using `Set` and it was an es6 addition. Do we need a polyfill? @mlkirkpatrick

mlkirkpatrick: It's to remove duplicates, but it won't remove nulls, so that's why the filter is there.

babel-polyfill should include it.

src/components/DynamicPages/RoomReservations/MeetingSpacesTable/index.js
pweikel: me neither :) fixed

src/components/DynamicPages/RoomReservations/MeetingSpacesTable/index.js
pweikel: Yeah. I updated it to the following:

```
 contactInfo: PropTypes.shape({
    status: PropTypes.string,
    contacts: PropTypes.arrayOf(PropTypes.shape({
      netid: PropTypes.string.isRequired,
    })),
  }),
```

src/tests/components/DynamicPages/RoomReservations/index.test.js
```diff
@@ -13,15 +12,17 @@ let props
...
-      fetchMeetingSpaces: jest.fn(),
       fetchSidebar: jest.fn(),
-      allMeetingSpacesStatus: 'true',
-      combinedStatus: 'true',
-      cfStaticStatus: 'true',
+      fetchContacts: jest.fn(),
+      fetchMeetingSpaces: jest.fn(),
+      allMeetingSpacesStatus: 'API_STATUS_NOT_FETCHED',
+      combinedStatus: 'API_STATUS_NOT_FETCHED',
+      cfStaticStatus: 'API_STATUS_NOT_FETCHED',
```
mlkirkpatrick, line 20: Please don't use a string literal. Import the constants.

src/tests/components/DynamicPages/RoomReservations/index.test.js
```diff
@@ -32,13 +33,20 @@ describe('components/DynamicPages/RoomReservations', () => {
...
 
   it('should call the fetch static content action with preview prop', () => {
+    enzymeWrapper.setProps({ pathSlug: 'slug', cfStaticStatus: 'true' })
     expect(props.fetchSidebar).toHaveBeenCalledWith('slug', true)
   })
 
+  it('should call the fetch contacts action', () => {
+    enzymeWrapper.setProps({ allMeetingSpacesStatus: 'API_STATUS_NOT_FETCHED' })
+    expect(props.fetchMeetingSpaces).toHaveBeenCalled()
+  })
```
mlkirkpatrick, line 48: This does not test the right thing. You need to check `props.fetchContacts` and set up the proper condition for it to be called.

src/tests/components/DynamicPages/RoomReservations/index.test.js
```diff
@@ -32,13 +33,20 @@ describe('components/DynamicPages/RoomReservations', () => {
   })
 
   it('should call the fetch meeting space action with preview prop', () => {
+    enzymeWrapper.setProps({ preview: true })
```
mlkirkpatrick, line 36: This is a bit of a cheat. Technically, I think the component should be refactored a tad. Really it should call the fetch action on mount, so setting props to force an update indicates that the component may not behave as intended.

It's common that these fetches need to happen both on mount _and_ on update (if a particular value changes.) To do this, you should make a separate function and put the fetch logic inside that, then just call it from both `componentDidMount` and `componentDidUpdate`. There are several examples of this throughout the code if you search for `checkFullyLoaded`.

src/tests/components/DynamicPages/RoomReservations/index.test.js
```diff
@@ -32,13 +33,20 @@ describe('components/DynamicPages/RoomReservations', () => {
...
     expect(props.fetchMeetingSpaces).toHaveBeenCalledWith(true)
   })
 
   it('should call the fetch static content action with preview prop', () => {
+    enzymeWrapper.setProps({ pathSlug: 'slug', cfStaticStatus: 'true' })
```
mlkirkpatrick, line 41: Setting the status to 'true' is a bit strange because that is not a value we ever expect the application to actually use. It's not the best representation of a real-world scenario.

src/tests/reducers/contacts.test.js
```diff
@@ -0,0 +1,38 @@
...
+      reducer(undefined, {
+        type: actions.RECEIVE_CONTACTS,
+        status: 'status from receiveContacts',
+        data: 'page from receiveContacts',
+      })
+    ).toEqual({
+      status: 'status from receiveContacts',
+      contacts: [],
+    })
+  })
```
mlkirkpatrick, line 37: This only checks a scenario where the data for RECEIVE_CONTACTS is empty or not in the structure expected by the reducer. It checks that the reducer can handle odd data without crashing, but does nothing to verify it can actually pass data through to the store correctly. You would need another test for that.

src/tests/components/DynamicPages/RoomReservations/index.test.js
pweikel: Updated to:
```
  allMeetingSpacesStatus: statuses.NOT_FETCHED,
      combinedStatus: statuses.NOT_FETCHED,
      cfStaticStatus: statuses.NOT_FETCHED,
```

src/tests/components/DynamicPages/RoomReservations/index.test.js
pweikel: Updated the test to the following:
```
  it('should call the fetch contacts action', () => {
    expect(props.fetchContacts).not.toHaveBeenCalled()
    enzymeWrapper.setProps({ allMeetingSpacesStatus: statuses.SUCCESS })
      const netids = ['one','three']
    expect(props.fetchContacts).toHaveBeenCalledWith(netids)
  })
```

src/tests/components/DynamicPages/RoomReservations/index.test.js
pweikel: Added fetches to `checkFullyLoaded()` and called it from both `componentDidMount` and `componentDidUpdate`

src/tests/components/DynamicPages/RoomReservations/index.test.js
pweikel: Removed the `setProps`

src/tests/reducers/contacts.test.js
pweikel: The test validates that the IDs are being created from the URL string:
```
it('should handle RECEIVE_CONTACTS', () => {
    expect(
      reducer(undefined, {
        type: actions.RECEIVE_CONTACTS,
        status: 'status from receiveContacts',
        data: {
          librarians: [
          {
            directoryUrl: 'https://library.nd.edu/directory/employees/one', 
          },
          {
            directoryUrl: 'https://library.nd.edu/directory/employees/two',
          }
        ]}
      })
    ).toEqual({
      status: 'status from receiveContacts',
      contacts: [
        {
          directoryUrl: 'https://library.nd.edu/directory/employees/one',
          netid: 'one'
        }, 
        {
          directoryUrl: 'https://library.nd.edu/directory/employees/two',
          netid: 'two'
        }
      ],
    })
  })
```

PR #943, opened 2020-10-19
replace facet ID space with underscore

pweikel:
Update to the Facet component to replace spaces with an underscore

### DIFF
--- a/src/components/Interactive/Facet/presenter.js
+++ b/src/components/Interactive/Facet/presenter.js
@@ -20,11 +20,11 @@ const Presenter = (props) => {
             <div key={option.key} className={styles.option}>
               <input
                 type='checkbox'
-                id={props.label + '_' + option.value.replace(/ /g,'_') }
+                id={props.label + '_' + option.value.replace(/ /g, '_')}
                 onChange={() => props.onFacetChange(option)}
                 checked={option.selected}
               />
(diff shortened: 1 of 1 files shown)


src/components/Interactive/Facet/presenter.js
```diff
@@ -20,11 +20,11 @@ const Presenter = (props) => {
             <div key={option.key} className={styles.option}>
               <input
                 type='checkbox'
-                id={props.label + ' ' + option.value}
+                id={props.label + '_' + option.value.replace(/ /g, '_')}
```
danthewolfe, line 23: I would recommend `/\s/g` instead of `/ /g`. It will match all whitespace characters, not just the literal unicode character 32. You can use `/[\s]+/` if you want to try to match multiple spaces to replace with a single underscore.

pweikel: Updated with "/\s/g". Great tip!

mlkirkpatrick (changes requested):
The facet label could also contain spaces, so make sure to replace spaces on that as well. (Put parens around the label and value to perform the replace on the concatenated string.)

danthewolfe (approved):

pweikel:
> The facet label could also contain spaces, so make sure to replace spaces on that as well. (Put parens around the label and value to perform the replace on the concatenated string.)

Good point! Updated it.

mlkirkpatrick (approved):

PR #949, opened 2020-11-12
LW-998 Remove redundancy with navigation model.

mlkirkpatrick:
There was an extra level of nested records that was completely unnecessary. ¯\\\_(ツ)_/¯

Refactor Section into its own file instead of cramming it in with the page presenter file.


danthewolfe (approved):
Assuming Travis, 👍

rdoughty (approved):

PR #1032, opened 2021-03-10
LW-1073 Add renewable column

pweikel:
Would appreciate some feedback on this. Especially on the efficacy of [this unit test](https://github.com/ndlib/usurper/pull/1032/files#diff-f47adfe15c9e458fb9ac568c8cdc3e22e3cdb2166af3b7d869f252a61d647e9eR59).


danthewolfe (approved):

src/components/Account/ResourceList/Resource/Cards/index.js
```diff
@@ -33,14 +34,18 @@ const Cards = (props) => {
...
         const className = (props.isMobileDetails ? styles.mobileDetail : styles[key]) +
           (props.listType === 'history' ? ` ${styles.circHist}` : '')
-        const value = (key === 'returnDate' && props.item.from === 'ILL')
-          ? 'Not Available'
-          : typy(props.item, key).safeString
-
+        let value = ''
+        if (key === 'renewable' && typy(props.item, key).isBoolean) {
+          value = props.item[key] ? 'Yes' : 'No'
+        } else {
```
mlkirkpatrick, line 43: The text "Yes" and "No" is a pretty generic string representation for a boolean; that's a good thing, but the way it's written here you won't immediately get that benefit if another boolean is added. There really isn't a reason to qualify it with `key === 'renewable'`, unless you want the value to be something special for this field which would _not_ apply to other booleans.

src/components/Account/ResourceList/Resource/Cards/index.js
```diff
@@ -33,14 +34,18 @@ const Cards = (props) => {
...
+          value = (key === 'returnDate' && props.item.from === 'ILL') ? 'Not Available' : typy(props.item, key).safeString
+        }
+        if (key === 'renewable' && props.item.from === 'ILL' && typy(props.item, key).isBoolean) {
+          value = props.item[key] ? <ToolTip value={'Yes'} /> : <ToolTip value={'No'} />
+        }
```
mlkirkpatrick, line 48: You _could_ set the value to a component here, but I wouldn't because it mixes the data type of `value` so that it is sometimes a string and sometimes a component. It also makes the previous code that sets the value to 'Yes' or 'No' irrelevant. Even if you did do it here, you wouldn't need to check if it is a boolean because you know that renewable **is** in fact a boolean.

I think it's more readable if you instead leave value as a simple string, and set another variable for the component it renders based on the key. It already does this with the "from" field, where it renders an <FromIcon> instead of the standard <span>; you just need one more if condition (or a switch case) to check if it's the "renewable" field, and if so render the ToolTip. (Also you would pass the value variable in to the prop, instead of hardcoding the yes and no a second time.)

src/components/Account/ResourceList/Resource/Cards/style.module.css
```diff
@@ -65,6 +65,11 @@
     -ms-grid-column: 2;
   }
 
+  .renewable{
+    -ms-grid-column: 4;
+    margin: auto;
+  }
+
```
mlkirkpatrick, line 72: Have you tested this on a small screen (mobile view)? I think some columns are hidden there so it might require a little more css than just this. I could be wrong though.

src/components/Account/ResourceList/style.module.css
```diff
@@ -115,7 +115,7 @@
 
 @media screen and (min-width: 62em) {
   .cardItem {
-    grid-template-columns: 2fr 1fr 1fr .5fr .25fr;
+    grid-template-columns: 2fr 1fr 1fr 1fr .75fr;
     -ms-grid-columns: 2fr 1fr 1fr .5fr .25fr;
```
mlkirkpatrick, line 119: You added a new column so this should be specifying six instead of five.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
```diff
@@ -51,7 +53,10 @@ describe('components/Account/ResourceList/Resource/Cards', () => {
...
           } else if (column === 'title') {
             expect(enzymeWrapper.find(TitleCard).exists()).toBe(true)
-          } else {
+          } else if (column === 'renewable') {
+            expect(enzymeWrapper.find(ToolTip).exists()).toBe(true)
+          }
+          if (column === 'renewable' && typeof props.item[column] === 'string') {
             const found = enzymeWrapper.findWhere(el => el.type() === Card && el.props().value === props.item[column])
             expect(found.exists()).toBe(true)
           }
```
mlkirkpatrick, line 62: This code block will never run because renewable is not a string and none of the other columns are called 'renewable'. I think this should just be an `else` so it will run for all columns except title and renewable.

src/tests/components/Account/ResourceList/Resource/ToolTip/index.test.js
```diff
@@ -0,0 +1,26 @@
...
+  })
+
+  it('should render Yes and tooltip in renewable column for ILL items', () => {
+    enzymeWrapper = setup({ value: 'Yes' })
+    expect(enzymeWrapper.containsMatchingElement(<div>Yes<span>Please click the &quot;view in ILL&quot; button to renew</span></div>)).toBe(true)
+  })
+  it('should render No and tooltip in renewable column for ILL items', () => {
+    enzymeWrapper = setup({ value: 'No' })
+    expect(enzymeWrapper.containsMatchingElement(<div>No<span>Renewal not available. Please return by due date.</span></div>)).toBe(true)
+  })
```
mlkirkpatrick, line 25: I don't like searching for exact text that is hardcoded because it often gets tweaked and it's not important to the test. Use `expect.any(String)` instead.

src/components/Account/ResourceList/Resource/ToolTip/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+import styles from './style.module.css'
+
+const ToolTip = (props) => {
+  const toolTip = (props.value === 'No' ? 'Renewal not available. Please return by due date.' : 'Please click the "view in ILL" button to renew')
```
mlkirkpatrick, line 7: Be consistent with punctuation; add a period to the end of the "Yes" message.

mlkirkpatrick (changes requested):

pweikel (commented):
I was going to reply to each comment directly but the option to do so doesn't seem to be working for some reason. I've addressed the feedback and update that PR.

src/tests/components/Account/ResourceList/Resource/ToolTip/index.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+  it('should render Yes and tooltip in renewable column for ILL items', () => {
+    enzymeWrapper = setup({ value: true })
+    expect.any(String)
+  })
+  it('should render No and tooltip in renewable column for ILL items', () => {
+    enzymeWrapper = setup({ value: false })
+    expect.any(String)
+  })
```
mlkirkpatrick, line 25: The description for these mention a "renewable column" and "ILL items." This happens to be the context in which it is used right now, but technically the ToolTip component doesn't know or care about any of these details. All it does is display a string appropriate for the value passed to it.

Additionally, these tests don't do anything. You're expecting a string but it's not even looking at the enzymeWrapper. You have to expect a string _inside of_ some element, whether that is a span, div, some other component, etc. For example:
```
const element = enzymeWrapper.find('span')
expect(element.exists()).toBe(true)
expect(element.text()).toEqual(expect.any(String))
```

Now along those lines, I think that you should test that the string for the true value and the false value are not the same string, otherwise you can't be sure that the value prop actually does anything; it might just be broken and it's always returning the false string, for example.

You can either do this by making your existing tests more specific so they actually do care about _part_ of the string, or you can make a thid test where you create two wrappers and just compare them to ensure the string inside them is not the same.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
```diff
@@ -51,7 +53,12 @@ describe('components/Account/ResourceList/Resource/Cards', () => {
             expect(enzymeWrapper.containsMatchingElement(find)).toBe(true)
           } else if (column === 'title') {
             expect(enzymeWrapper.find(TitleCard).exists()).toBe(true)
-          } else {
+          } else if (column === 'returnDate') {
+            const find = <span>Not Available</span>
+            expect(enzymeWrapper.containsMatchingElement(find)).toBe(true)
+          } else if (column === 'renewable') {
+            expect(enzymeWrapper.find(ToolTip).exists()).toBe(true)
+          } else if (column !== 'renewable' || 'title') {
```
mlkirkpatrick, line 61: The "Not Available" is a byproduct of changing the type to ILL; there are already test cases specifically for ILL items, so you shouldn't be changing this one from NDU. By doing this, you are no longer testing NDU items and whether they get a return date or not.

Also, the last if condition is unnecessary and incorrect. We're already chaining conditions, so you don't need to check if it's not renewable because the check for if it is renewable is above. You also can't do a not comparison with an or operator like that; this would return true if the column value was 'title' because the string 'title' is truthy. What you meant was `column !== 'renewable' && column !== 'title'`, but I just explained that the whole if condition is unnecessary anyway. :)

mlkirkpatrick (changes requested):

src/tests/components/Account/ResourceList/Resource/ToolTip/index.test.js
pweikel: Okay. I follow you. I've updated the the test titles to be more generic and the parameters to be more or less specific.

src/components/Account/ResourceList/Resource/Cards/index.js
```diff
@@ -33,14 +34,18 @@ const Cards = (props) => {
...
-          : typy(props.item, key).safeString
-
+        let value = ''
+        if (typy(props.item, key).isBoolean) {
+          value = props.item[key] ? 'Yes' : 'No'
+        } else {
+          value = (key === 'returnDate' && props.item.from === 'ILL') ? 'Not Available' : typy(props.item, key).safeString
+        }
+        if (key === 'renewable' && props.item.from === 'ILL') {
+          return <ToolTip key={key} value={props.item[key]} />
```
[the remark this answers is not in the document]
pweikel: I took your advice and revised this to pass the prop value prop to the component

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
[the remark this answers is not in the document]
pweikel: `          } else if (column !== 'renewable') {
            const found = enzymeWrapper.findWhere(el => el.type() === Card && el.props().value === props.item[column])
            expect(found.exists()).toBe(true)`
Thank you for explaining the logic there. I believe that I modified the unit test so that it looks at the columns that have not yet been reviewed.

src/tests/components/Account/ResourceList/Resource/ToolTip/index.test.js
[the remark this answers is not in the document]
pweikel: Updated the unit test to look for part of the string instead of the whole string.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
```diff
@@ -51,7 +53,9 @@ describe('components/Account/ResourceList/Resource/Cards', () => {
             expect(enzymeWrapper.containsMatchingElement(find)).toBe(true)
           } else if (column === 'title') {
             expect(enzymeWrapper.find(TitleCard).exists()).toBe(true)
-          } else {
+          } else if (column === 'renewable' && props.item.from === 'ILL') {
+            expect(enzymeWrapper.find(ToolTip).exists()).toBe(true)
```
mlkirkpatrick, line 57: This will never execute since the test case is for an item from NDU, not ILL. You need to make a new test case with an ILL item and check for the ToolTip.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
```diff
@@ -51,7 +53,9 @@ describe('components/Account/ResourceList/Resource/Cards', () => {
...
+          } else if (column !== 'renewable') {
             const found = enzymeWrapper.findWhere(el => el.type() === Card && el.props().value === props.item[column])
             expect(found.exists()).toBe(true)
```
mlkirkpatrick, line 60: What about testing the renewable column? We should expect that true or false get converted to "Yes" and "No", so I would at least check that the renewable column displays _some_ string and not just an empty cell, to ensure the conversion from bool to string works.

src/components/Account/ResourceList/Resource/Cards/index.js
```diff
@@ -33,14 +34,18 @@ const Cards = (props) => {
...
-          : typy(props.item, key).safeString
-
+        let value = ''
+        if (typy(props.item, key).isBoolean) {
+          value = props.item[key] ? 'Yes' : 'No'
+        } else {
+          value = (key === 'returnDate' && props.item.from === 'ILL') ? 'Not Available' : typy(props.item, key).safeString
+        }
+        if (key === 'renewable' && props.item.from === 'ILL') {
+          return <ToolTip key={key} title={props.item.title} value={props.item[key]} />
```
mlkirkpatrick, line 47: Just to point out: `key` isn't needed as a prop here, but it's not doing any harm either. You typically only need this when looping through an array.

mlkirkpatrick (changes requested):
Close. :) Just a couple more changes to the unit tests and then this looks good.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
pweikel: Got it. I added a test to look for the string in the renewable column.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
pweikel: Okay. I made a[ separate test](https://github.com/ndlib/usurper/pull/1032/files#diff-f47adfe15c9e458fb9ac568c8cdc3e22e3cdb2166af3b7d869f252a61d647e9eR95) for the ToolTip.

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
```diff
@@ -53,11 +53,12 @@ describe('components/Account/ResourceList/Resource/Cards', () => {
...
           } else if (column === 'title') {
             expect(enzymeWrapper.find(TitleCard).exists()).toBe(true)
-          } else if (column === 'renewable' && props.item.from === 'ILL') {
-            expect(enzymeWrapper.find(ToolTip).exists()).toBe(true)
           } else if (column !== 'renewable') {
             const found = enzymeWrapper.findWhere(el => el.type() === Card && el.props().value === props.item[column])
             expect(found.exists()).toBe(true)
+          } else if (column === 'renewable') {
+            const found = enzymeWrapper.findWhere(el => el.type() === Card && el.hasClass('renewable'))
+            expect(found.contains(<span className="detailValue">No</span>)).toBe(true)
```
mlkirkpatrick, line 61: I'd prefer not including the `className="detailValue"` on this. You can test just the text with:
```
expect(found.text()).toEqual('No')
```

mlkirkpatrick (approved):

mlkirkpatrick (approved):

src/tests/components/Account/ResourceList/Resource/Cards/index.test.js
pweikel: changed to: `expect(found.containsMatchingElement(<span>{expect.any(String)}</span>)).toBe(true)`

PR #1045, opened 2021-03-24
Add renewable column

pweikel:


src/components/Account/ResourceList/Resource/ToolTip/index.js
```diff
@@ -4,18 +4,17 @@ import PropTypes from 'prop-types'
 import styles from './style.module.css'
 
 const ToolTip = (props) => {
```
pweikel, line 6: Do you think I should move the `ToolTip` component into `src/components` or leave it where it's at?

danthewolfe (approved):

src/components/Account/ResourceList/Resource/ToolTip/index.js
danthewolfe: If you think it is likely to be reused elsewhere (or already is), you may want to move it higher in the tree to `src/components`, otherwise I think it is likely fine where it is.

mlkirkpatrick: I would move it to `src/components/Interactive`

src/components/Account/ResourceList/Resource/ToolTip/index.js
```diff
@@ -4,18 +4,17 @@ import PropTypes from 'prop-types'
...
+    <div aria-labelledby={props.title} className={styles.tooltip}>{props.value}
+      <span role='tooltip' id={props.title} className={styles.tooltiptext}>{props.message}</span>
     </div>
   )
 }
 
 ToolTip.propTypes = {
   value: PropTypes.bool,
   title: PropTypes.string,
+  message: PropTypes.string,
```
mlkirkpatrick, line 17: I might rename these props to try to be less ambiguous. It also doesn't seem like "title" is really necessary as a prop. Of course, ids are useful for accessibility but are a bit of a sore-spot for React since it goes against the idea of reuse. But ultimately we don't care what the id is, just that it is unique. I'd be tempted to just generate a random string (long enough to be unlikely to produce duplicates) for the id, removing the need to pass it down as a prop altogether.

You don't have to do that, but in any case, I think the names could use some refining.

Futhermore, it stands to reason that these props should be required.

mlkirkpatrick (approved):

src/components/Account/ResourceList/Resource/ToolTip/index.js
pweikel: Made a `makeId` function that I added to HelperFunctions. This provides both a `key` and an `ID` for the component.

src/components/Account/ResourceList/Resource/ToolTip/index.js
pweikel: added to `src/components/Interactive`

src/components/Interactive/ToolTip/index.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { makeId } from 'constants/HelperFunctions'
+
+import styles from './style.module.css'
+
+const ToolTip = (props) => {
+  const id = makeId(15)
+  const key = makeId(15)
```
mlkirkpatrick, line 9: key is not needed

src/tests/components/Interactive/ToolTip/index.test.js
```diff
@@ -16,15 +16,9 @@ describe('components/Account/ResourceList/Resource/ToolTip', () => {
   })
 
   it('should render a tooltip ', () => {
-    enzymeWrapper = setup({ value: true })
+    enzymeWrapper = setup({ value: 'Yes', message: 'Please click the "view in ILL" button to renew.'})
     const element = enzymeWrapper.find('span')
     expect(element.exists()).toBe(true)
     expect(element.text()).toContain('ILL')
```
mlkirkpatrick, line 22: Now that you are passing the tooltip text in as a prop, you could check that they exactly match.

mlkirkpatrick (approved):

src/components/Interactive/ToolTip/index.js
pweikel: Right. I was getting a console warning but it was for `cards/index.js`. Removed key from ToolTip

PR #1047, opened 2021-03-24
added showDescription to items-request page

pweikel:
Melissa's request for prep to do some "experimenting"


src/components/Contentful/StaticContent/Sidebar/presenter.js
```diff
@@ -11,7 +11,7 @@ import ServicePoint from 'components/Contentful/ServicePoint'
 const Presenter = ({ cfStatic, inline, showDescription }) => (
   <aside key={`ContentfulSidebar_${cfStatic.sys.id}`} className={inline ? '' : 'col-md-4 col-sm-5 col-xs-12 right'}>
     <PageLink className='button callout' cfPage={cfStatic.fields.callOutLink} />
-    {showDescription ? <LibMarkdown>{cfStatic.fields.shortDescription}</LibMarkdown> : null}
+    {(window.location.pathname === '/items-requests' && showDescription) ? <LibMarkdown>{cfStatic.fields.shortDescription}</LibMarkdown> : null}
```
mlkirkpatrick, line 14: You shouldn't be checking for the path like this. Just pass showDescription=true when it's relevant and let that be that.

src/components/Account/AccountPageWrapper/presenter.js
```diff
@@ -42,7 +42,7 @@ const AccountPageWrapper = (props) => {
         { props.customSidebar ? (
           props.customSidebar
         ) : (
-          <StaticSidebar slug={props.slug} preview={props.preview} />
+          <StaticSidebar slug={props.slug} preview={props.preview} showDescription />
```
mlkirkpatrick, line 45: This would cause all account pages to show the description. If you only want it to do so for Items & Requests, you'll need to add a prop here and pass it to the AccountPageWrapper only from that one page.

mlkirkpatrick (changes requested):

src/components/Contentful/StaticContent/Sidebar/presenter.js
pweikel: added prop to items-requests page to activate showDescription

src/tests/components/Contentful/StaticContent/Sidebar/presenter.test.js
```diff
@@ -45,11 +45,6 @@ describe('components/Contentful/StaticContent/Sidebar/presenter', () => {
     props = undefined
   })
 
-  it('should show shortDescription if provided', () => {
-    const found = enzymeWrapper.findWhere(el => el.text() === props.cfStatic.fields.shortDescription)
-    expect(found.exists()).toBe(true)
-  })
```
mlkirkpatrick, line 51: Why remove this test? Having a unit test to check that it does show up is a good idea.

pweikel: Added unit test back in

mlkirkpatrick (approved):

danthewolfe (approved):

PR #1063, opened 2021-04-19
updated alert border colors

pweikel:


danthewolfe (approved):

src/components/Contentful/Alert/style.module.css
```diff
@@ -8,13 +8,23 @@
...
+}
+.alertSection.warning {
+  border: solid 2px #ad5458;
+}
+.alertSection.serviceAvailable {
+  border: solid 2px #1d7a4f;
+}
+.alertSection.serviceLimited {
+  border: solid 2px #1d7a4f;
+}
```
mlkirkpatrick, line 25: Combine rules that use the same styles like so:
```
.alertSection.serviceAvailable,
.alertSection.serviceLimited {
  border: solid 2px #1d7a4f;
}
```

src/components/Contentful/Alert/style.module.css
```diff
@@ -8,13 +8,23 @@
...
 .alertSection.serviceUnavailable {
-  border: solid 2px #AA272F;
+  border: solid 2px #1d7a4f;
 }
```
mlkirkpatrick, line 28: This isn't in use, but if it was, it should use the same color as warning. Suggest combining those two.

mlkirkpatrick (approved):
Approved, with minor suggested changes

src/components/Contentful/Alert/style.module.css
pweikel: Condensed CSS

src/components/Contentful/Alert/style.module.css
pweikel: Combined

PR #1080, opened 2021-05-17
Update to Facet component

pweikel:
Updated the Facet component so that it correctly identifies the label


src/components/Interactive/Facet/presenter.js
```diff
@@ -25,9 +25,9 @@ const Presenter = (props) => {
                 checked={option.selected}
               />
               <label htmlFor={(props.label + '_' + option.value).replace(/\s/g, '_')}>
-                <button className={'custom-style link-like ' + styles.linkText} tabIndex={0} onClick={() => props.onFacetChange(option)}>
+                <div aria-label={props.label} className={'custom-style link-like ' + styles.linkText} tabIndex={0} onClick={() => props.onFacetChange(option)}>
```
mlkirkpatrick, line 28: I don't think that label is quite what you want. That is just the name of the facet category, not the option itself.

mlkirkpatrick (changes requested):

src/components/Interactive/Facet/presenter.js
pweikel: Good point.
Changed to: `aria-label={('Checkbox for ' + option.value)} `

mlkirkpatrick (approved):

danthewolfe (approved):

PR #1081, opened 2021-05-17
update Current Hours component

pweikel:
updated component so that the headings are in the correct order


danthewolfe (approved):

mlkirkpatrick (changes requested):
Needs a style change so that text will not be misaligned between the label and the hours. Should just be `line-height: 1.5em;`

mlkirkpatrick (approved):

PR #1092, opened 2021-06-10
fix for label

pweikel:


src/components/Interactive/Facet/presenter.js
```diff
@@ -25,7 +25,7 @@ const Presenter = (props) => {
                 checked={option.selected}
               />
               <label htmlFor={(props.label + '_' + option.value).replace(/\s/g, '_')}>
-                <div aria-label={('Checkbox for ' + option.value)} className={'custom-style link-like ' + styles.linkText} tabIndex={0} onClick={() => props.onFacetChange(option)}>
+                <div aria-label={('Checkbox for ' + option.value)} className={'custom-style link-like ' + styles.linkText} tabIndex={0}>
```
mlkirkpatrick, line 28: Keyboard users can still tab to this element, but won't be able to activate it which is confusing because it looks clickable. Either remove tabIndex so tabbing will only highlight the checkboxes or other changes are needed.

mlkirkpatrick (changes requested):

src/components/Interactive/Facet/presenter.js
pweikel: Good point! Removed it

mlkirkpatrick (approved):

danthewolfe (approved):

PR #1107, opened 2021-08-13
Fix clicking on facet label

mlkirkpatrick:
This issue was fixed in prep a few months ago but never got merged to master.


pweikel (approved):
totally approved!

PR #1114, opened 2021-09-07
LW-1222 Adding www to library.nd.edu

danthewolfe:
* Added `https://www.library.nd.edu` to arrays that determine if links are "local"
* Changed `https://library.nd.edu` to `https://www.library.nd.edu`


src/components/Interactive/Link/index.js
```diff
@@ -105,10 +105,11 @@ export const LibLink = (props) => {
     'http://alpha.library.nd.edu',
     'https://beta.library.nd.edu',
     'http://beta.library.nd.edu',
-    'https://library.nd.edu',
+    'https://www.library.nd.edu',
```
mlkirkpatrick, line 108: We should keep the existing one just in case and append the https://www to the list.

mlkirkpatrick (changes requested):

src/components/Interactive/Link/index.js
mlkirkpatrick: I fixed it

mlkirkpatrick (approved):

PR #1119, opened 2021-09-16
added Canonical component and unit test

pweikel:
Added Canonical url to the Events and unit test


mlkirkpatrick (approved):

pweikel:
> This might be backwards. Rather than having the event page declare that it is canonical, I think you are supposed to have duplicates point to the canonical page. (E.g. The EventCards should have a rel="canonical" pointing to the event page.)

That was my initial approach until I discovered that unless the **rel="canonical"** is in the <head /> of the document, it is otherwise ignored by search engines. There was some debate about the usefulness of self-referencing canonical URLs but Google's John Mueller recently indicted that they do in fact help but are not critical to search performance.

mlkirkpatrick:
That's fine, but it still doesn't address the need for a duplicate (I.E. event cards on the home page) to point to the canonical version (the event page.) Both should be used together, as I understand it.

danthewolfe:
> That's fine, but it still doesn't address the need for a duplicate to point to the canonical version (the event page.) Both should be used together, as I understand it.

I think what we want is for the events that are marked up as structured data on the homepage and the events page(s) need to point to the actual individual events pages.

mlkirkpatrick:
Like Dan mentioned, it might not address the issue this ticket was created for directly, but it does seem logical for the structured data on the card to point to the event's page. I'm thinking the appropriate attribute would be [sameAs](https://schema.org/Thing), but I'll leave that up to @pweikel

pweikel:
> Like Dan mentioned, it might not address the issue this ticket was created for directly, but it does seem logical for the structured data on the card to point to the event's page. I'm thinking the appropriate attribute would be [sameAs](https://schema.org/Thing), but I'll leave that up to @pweikel

I agree with your assessment here of using `sameAs` to reference the event page. The original impetus of the ticket was to eliminate the duplicate events showing up on the SERP. This might help, but, I don't think it will fully resolve the issue. The only way to know will be to add the property and see what happens after Google crawls it.

danthewolfe (approved):

mlkirkpatrick (approved):

PR #1120, opened 2021-09-16
updated Event schema

pweikel:
Updated some fields within the Event schema


src/components/Contentful/Event/presenter.js
```diff
@@ -35,9 +35,10 @@ const PagePresenter = ({ entry }) => {
       itemType='http://schema.org/Event'
       itemProp='mainEntity'
     >
-      {entry.shortDescription && (<meta itemProp='description' content={entry.shortDescription} />) }
+      {entry.shortDescription && (<meta itemProp='about' content={entry.shortDescription} />) }
       <meta itemProp='startDate' content={entry.startDate} />
       <meta itemProp='endDate' content={entry.endDate} />
+      <meta itemProp='audience' content={entry.audience} />
```
mlkirkpatrick, line 41: I don't think this field will work very well as-is. `entry.audience` is an array of strings. The schema probably expects a structured object and a single string for the audience name.

mlkirkpatrick (changes requested):

src/components/Contentful/Event/presenter.js
pweikel: You're right about this. It does appear that the schema expects a single string so this data point doesn't look like it will work for this purpose. If it accepted multiple strings, we could have used it, but that's not the case.

src/components/Contentful/Event/presenter.js
```diff
@@ -82,9 +82,8 @@ const PagePresenter = ({ entry }) => {
             </div>
           </div>
           <LibMarkdown itemProp='description'>{ body }</LibMarkdown>
-          <meta itemProp='sponsor' content={entry.sponsors} />
           <Related className='p-resources' title='Resources' showImages={false}>{ entry.relatedResources }</Related>
-          <Sponsorships sponsors={entry.sponsors} />
+          <Sponsorships itemProp='sponsor' content={entry.sponsors} />
```
mlkirkpatrick, line 85: `itemProp` needs to go on an HTML element, not the React component. If you put it here, it just passes as a prop, and then the React component needs to pass it down further. You might as well build it right into the component to include the itemProp on a div.

src/components/Sponsorships/index.js
```diff
@@ -2,20 +2,20 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import LibMarkdown from 'components/LibMarkdown'
 
-const Sponsorships = ({ sponsors }) => {
-  if (!sponsors) {
+const Sponsorships = ({ content }) => {
+  if (!content) {
```
mlkirkpatrick, line 6: I'm not sure why you changed the name of this prop. It seems like maybe you were going to pass the meta tag inside it? I don't think that's necessary, as you can just include the meta tag directly in the rendered output instead of needing to pass it in.

mlkirkpatrick (changes requested):

src/components/Contentful/Event/presenter.js
```diff
@@ -80,7 +81,8 @@ const PagePresenter = ({ entry }) => {
               />
             </div>
           </div>
-          <LibMarkdown itemProp='articleBody'>{ body }</LibMarkdown>
+          <LibMarkdown itemProp='description'>{ body }</LibMarkdown>
+          <meta itemProp='sponsor' content={entry.sponsors} />
```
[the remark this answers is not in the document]
pweikel: Good call. I updated the `Sponsorships` component to use the content property so that it is compatible with the schema property. Also, I noticed the `Sponsorships` component didn't have unit tests so I added those.

src/components/Contentful/Event/presenter.js
pweikel: I follow you. I was thinking that it would best to have all the props for Event in the same component, but now I see that it will still be within the scope of the Event schema even tho it is rendered in the Sponsorships component.

src/components/Sponsorships/index.js
pweikel: Acting your previous advice, I've updated the location of the itemProp into the Sponsorships component and reverted the name of this prop.

mlkirkpatrick (approved):

mlkirkpatrick (approved):

danthewolfe (approved):

PR #1176, opened 2022-06-08
added meta description tag

pweikel:


src/components/OpenGraph/index.js
```diff
@@ -15,6 +15,7 @@ const OpenGraph = (props) => {
 
   return (
     <Helmet>
+      <meta name='description' content={description} />
```
danthewolfe, line 18: I know we have gotten Google Search Console complaints about certain fields being too long before. I don't think there's a specific hard length. See: https://moz.com/learn/seo/meta-description#:~:text=Meta%20descriptions%20can%20technically%20be,between%2050%20and%20160%20characters.

Maybe we should truncate the description like this:
```
<meta name='description' content={description.substring(0, 150)} />
```

pweikel: Looks like the latest thinking is that 160 is the optimal character length because of a change in the size of the SERP page so I will adjust to reflect based on [this](https://seopressor.com/blog/google-title-meta-descriptions-length/) article updated: 2022

danthewolfe (approved):

